PR #21, opened 2016-07-04
Satoshi's initial was wrong

girino:


jcvernaleo:
Good catch.  Thanks.
OK

PR #25, opened 2016-08-15
Add documentation for stake pool mining with Paymetheus.

jrick:
No idea if the images are in the correct location for the site.


jrick:
Updated with fixes from @jolan

jolan:
Works now and looks good, thanks.

PR #27, opened 2017-01-09
Update linux.md

NUKE1989:
Fix dcrctl command example for linux user-guide

### DIFF
--- a/docs/getting-started/user-guides/linux.md
+++ b/docs/getting-started/user-guides/linux.md
@@ -248,7 +248,7 @@ Decred. Open a third terminal and go to your Decred directory. Type in
 (once again using your values for the square brackets):
 
 ```no-highlight
-~/Decred/$ ./dcrctl -u [daemon username] -P [daemon password] --wallet walletpassphrase [wallet password] 0
+~/Decred/$ ./dcrctl -u [wallet username] -P [wallet password] --wallet walletpassphrase [wallet password] 0
 ```
 
 What we are doing here is authenticating with the daemon and using the


docs/getting-started/user-guides/linux.md
```diff
@@ -248,7 +248,7 @@ Decred. Open a third terminal and go to your Decred directory. Type in
 (once again using your values for the square brackets):
 
 ```no-highlight
-~/Decred/$ ./dcrctl -u [daemon username] -P [daemon password] --wallet walletpassphrase [wallet password] 0
+~/Decred/$ ./dcrctl -u [wallet username] -P [wallet password] --wallet walletpassphrase [wallet password] 0
 ```
 
 What we are doing here is authenticating with the daemon and using the
```
jcvernaleo: If you changed daemon to wallet above I think you need to change daemon to wallet in this line as well.

jcvernaleo: Also could you use a more descriptive commit message?

Something like "Fix dcrctl command example for linux user-guide" or anything that tells use what is in the commit/

jcvernaleo:
@NUKE1989 Thanks!  Two minor points and then I can get this in.
With the docs site it is necessary to regenerate the web files after making changes using the command
```
mkdocs build
```
as the README says but I can do that in a followup commit if you don't have mkdocs installed.

docs/getting-started/user-guides/linux.md
NUKE1989: yeah you are correct. 
Also i dont know how github works so i will now know better for next time.

jcvernaleo:
So you can just make a new commit with that 2nd change (making sure to give it a better commit message), push that commit, and then if it looks good I can take care of the rest.

NUKE1989:
ok, so i dont want to waste your time. I certainly need to learn more about how github works. So go ahead and just fix the error. But, while i agreed that that i also need to change the line you said, now that i look at it, i am not sure i understand. Wasnt that the line i actually changed? I cant understand what is the difference.
Also you mean that i should close this pull request, do what must be done, and open a new one?

Sorry for the questions. I am not familiar with github. I will try to find material to understand how it works.

jcvernaleo:
Not a problem.  I'll make the change and give you credit.

Issue #49, opened 2017-01-31
Remove external js/css dependancies


jcvernaleo:
There are two bits of js and css that are loaded externally.  Those should be moved into the repo and self-hosted.

jcvernaleo:
With #50 in, only MathJax remains not local.  That one seems a big harder to do locally though.

jholdstock:
I spent some time looking at this. The difficulty in moving MathJax to be hosted alongside the rest of the docs is that the server software used by dcrdocs (tornado) attempts to insert a `livereload.js` script into the MathJax files.

This can be recreated by attempting to host the file `MathJax/extensions/MathMenu.js` in tornado server. This file contains the string "</title>" and this is the point where tornado attempts to insert `livereload.js`

jcvernaleo:
So it sounds to me like this is something we have to close and mark as 'won't fix' since I do not see us wanting to get into hacking tornado.

jcvernaleo:
@jholdstock Are you ready to accept defeat on this one like I did? :)

If so I'd like to close it.

jholdstock:
Sure - go ahead and shut it down

jcvernaleo:
Since @jolan, me, and @jholdstock have all tried (very little alphabetic diversity there :) ) and each decided this can't be done in a useful manner, closing this issue.

PR #50, opened 2017-01-31
Use Local font-awesome.

jcvernaleo:
This requires an extra install with pip:
pip install fontawesome-markdown

Partially takes care of #49


jcvernaleo:
I wasn't thrilled that it used an older version myself but figured it was worth it for one less network dependance.  I guess we could do a PR on the upstream to try and get them to upgrade but I'm not sure it matters enough to me.

PR #63, opened 2017-02-11
Newdesign rebase

gozart1:
@jcvernaleo I've gone ahead and rebased #53  
@karamble lmk if you'd prefer to squash it yourself since most of these commits are yours, the conflicts where simple to deal with. I did some sanity checking and the rebase seems to have gone smoothly.

PR #53, opened 2017-02-04
New Design Update HTML, JS, CSS, IMG and ICONS

karamble:
#52 This is the initial commit with the HTML,JS, CSS, IMG and ICONS 
contents submitted by Lunitee

This theme serves all those needed assets locally and is using bootstrap, js and webfonts.

Site can be browsed with enabled NoScript with same user experience.
Custom Webfonts and Icons won't be displayed in this case.

You can enable the theme by editing the **mkdocs.yml** file and point to the template directory:
`
theme: null
`
`
theme_dir: tpldcrdocs/
`
The only relevant files for this PR are in the tpldcrdocs/ directoy

PR #72, opened 2017-02-14
New design based on the master repo

### LINKED ISSUE #52: Custom Theme with new Design
We need to update the theme according to the new Design specifics.

You can find informations on how to customize themes for mkdocs on their user-guide at
http://www.mkdocs.org/user-guide/custom-themes/

Therefore the template have to be created into a single directory which contains all needed assets, HTML, CSS, JS and FONTS. This directory should be created next to the mkdocs.yml configuration file in the document root.

The requirement is to disclaim JS if possible. And if JS have to be used it have to be served locally and the Sites usablility should be ensured, even when NoScript browsing is enabled.

You can activate the new template by editing the mkdocs.yml and point to the directory with the following parameters:
`
theme: null
theme_dir: YOU_THEME_DIRECTORY/
`

karamble:
closes #52 
added directory tpldcrdocs/ with contents of the new design

mkdocs.yml configuration files was edited to make the new design the default


jcvernaleo:
@karamble how does this related to #53 ?

karamble:
#53 can be removed, because i screwed the #53 PR - it was not mergeable without conflicts so i redit the work

jcvernaleo:
@karamble in the future, do not open new PRs for the same work.  I can help you either help  you fix your branch or replace it with a new one on the same PR.  I've seen some truly amazing messes with git and it is always fixable.

jolan:
#63 can be removed too.  Nothing like 3 PRs for the same thing :-1: :laughing:

jolan:
Looks good to me as well.

PR #74, opened 2017-02-14
Update obtaining-dcr.md

jzbz:
Ranked exchanges by volume.
Removed YoBit, they do not allow users to withdraw DCR, their wallet is broken.
Removed C-cex, they don't list Decred.

Better to steer users towards exchanges that have liquidity and will allow them to deposit/withdraw with minimum hassle.

### DIFF
--- a/docs/getting-started/obtaining-dcr.md
+++ b/docs/getting-started/obtaining-dcr.md
@@ -30,14 +30,12 @@ well documented in the
 ### **<i class="fa fa-exchange"></i> Exchanges**
 
 Since the release of Decred on February 8th, 2016, there are five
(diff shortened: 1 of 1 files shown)


docs/getting-started/obtaining-dcr.md
```diff
@@ -30,14 +30,12 @@ well documented in the
 ### **<i class="fa fa-exchange"></i> Exchanges**
 
 Since the release of Decred on February 8th, 2016, there are five
```
alexlyp: *five should be three now

jzbz:
Good catch. Fixed.

PR #96, opened 2017-03-03
introducing Multilanguage with build_script.sh

karamble:
I have constructed a new directory list and dublicated the mkdocs.yml manifests so we are able to introduce multilanguage support.

Additionaly there is a build_docs.sh script which builds the various language versions into different directorys. Resulting into following directory structure currently:

`en_site/
ru_site/
de_site/`

The Serveradmin of the webserver would have to point the various language subdomains (ru.docs.decred.org) into the specific language_site directories. Let me know what you think!


jcvernaleo:
Could you edit the title of this to make it clear that this is still a work in progress and not ready to be fully reviewed/merged?

karamble:
Ok now we discarded to use subdomain structure for the different languages and opted in for using language subdirectory folders.

These changes are now reflected in the build_docs.sh script.

If you want to use the `mkdocs serve` functionality to serve the docs locally without the need to build you should add the parameter to select which language to serve you are currently working on:
`mkdocs serve -f en_mkdocs.yml `

karamble:
I have rebased to the latest version.

Our consensus was that we will use subdirectory structure for the different languages.
In this PR prebuild there are the docs structures for /ru and /de as additional languages.

build-docs.sh script should be used to build the complete documentation in the various languages. As result you will get a site/ directory with the default language en build out from docs/ directory.

Additionaly it builds site/ru/ and site/de/ from ru_docs/ and de_docs/ directories.

No duplicates of css, js and Icons directories, the build-docs.sh script takes care of removing those directories after build and adds Symlinks for those dirs.

The only thing that currently is duplicated are site/mkdocs/ directory which holds the search.js functionality.

The second duplicate is the /img/ directory, i don't know how this should behave over the various languages, i suggest keeping the /img/ folder in the document root so the documentations can share images, if a translation needs specific languages they can put their versions in their respective subdirectories and linking them proper in the .md files. No practical experience on my site with writing the docs so far.

jcvernaleo:
@dajohi How does this look to you?

jcvernaleo:
@karamble did you ever get the code in to copy new files to the translated directories?

Currently, this does not like to the translations anywhere, correct?  I'm curious because if it does not link to them from the english site we can probably get this in and then get the translations going as a separate PR and we could also worry about how to show the translated pages separately as well.

jcvernaleo:
As for your img folder idea (having it in the document root so it is shared and then only ones that need translating going into the language dirs, I think that sounds great.

karamble:
There are currently de and ru versions implemented into the build-script. After the first translation is done we should add a language navigation on the website. i suggest using the famfam flag icons to integrate a language navigation. http://www.famfamfam.com/lab/icons/flags/

jcvernaleo:
Going to run it by @dajohi before merging it though in case he has any thoughts/issues.

jcvernaleo:
@dajohi said on irc that he is cool with this going in as long as it has a script for him (which it does) so in it goes.

Thanks for all the work on it @karamble

PR #81, opened 2017-02-20
Dark theme, get rid of top bar

gozart1:
https://github.com/decred/dcrdocs/issues/73

Stuff in this PR:
- Dark theme for main content
- Unify navigation/search markup into a single template
- Each page links to github page of current content
- Refactor css a bit ( hence the 877 deletions )

PR #99, opened 2017-03-05
refactor html/css, remove top bar, link to current github page

gozart1:
This PR replaces #81

Felt that I should decouple these changes from light vs dark styling.

- unify duplicate navigation/search markup into a single template
- links to currently viewed content to its github page
- removes top bar
- various css tweaks ( eg improve spacing for mobile reading ) 

preview here
`http://dcrdocs.netlify.com/`


jcvernaleo:
I agree, splitting the dark theme out into something separate is probably best (and it means this can go in now).

PR #111, opened 2017-03-17
Ensure table text and list text are the same size as other text

### LINKED ISSUE #104: CSS: Lists are different font/style from rest of page.
screen shot 2017-03-10 at 12 05 26 am

Also, 3 is within `<p>` tags inside `<li>` tags, while 1 and 2 are simply inside `<li>` tags. These have two different styles (I also don't know why the markdown is being parsed differently with this list).

jholdstock:
Fixes #104

### DIFF
--- a/tpldcrdocs/css/base.css
+++ b/tpldcrdocs/css/base.css
@@ -200,7 +200,6 @@ div[role=main] ul {
 }
 div[role=main] li {
     font-family: 'Source Sans Pro', sans-serif;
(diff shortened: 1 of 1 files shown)


jcvernaleo:
Looks fine to me.  Since it is design related I'm going to solicit designer opinion on slack just to be sure before approving and merging it.

fai04d:
Yeah, fine with me.

PR #134, opened 2017-03-24
Seed reuse warning

### LINKED ISSUE #108: warn about seed re-use/sharing
Numerous support issues have been related to plugging the same seed into multiple wallets, web wallet, etc. and using them concurrently.

This is unsupported and causes problems when the wallets fail to receive/see all transactions/addresses.

Should strongly recommend "new wallet, new seed".

jholdstock:
Closes #108


jholdstock:
Fixed the links. That was a dumb one, especially when the console output explains it so clearly. The kind of mistake you only make once :slightly_smiling_face:

jcvernaleo:
Great, that takes care of the issues I had.

PR #120, opened 2017-03-20
Added proper installation and continuation links to the golden path

fai04d:
Resolves #118 

Added links and continuation for Decrediton users.

PR #145, opened 2017-03-29
Updated Decrediton link.

fai04d:
This was supposed to be an update to #120 to fix the above link, but due to a conflict I tried to delete the branch and start a new one with the same name. Github didn't like that and closed the PR and I can't reopen it, hence this new PR.


jcvernaleo:
Usually github is very good about that but occasionally you hit a weird spot with it and that is always super annoying.

PR #166, opened 2017-04-05
Add using-testnet.md

### LINKED ISSUE #78: Create Running A Testnet Node Guide
Create guide for running a node on testnet.

This should be included in #77

gratefulcheddar:
This PR adds the Using Testnet Guide (/getting-started/using-testnet.md), and recommends using the testnet to beginners on the beginner-guide.md page.

Because testnet2 currently requires compiling from source, I think this PR should be held until new binaries are released, but comments, criticisms, etc,.. much appreciated.  

Closes #78

### DIFF
--- a/docs/getting-started/using-testnet.md
+++ b/docs/getting-started/using-testnet.md
@@ -1,12 +1,12 @@
 # Using Testnet Guide
 
-last updated for v1.0.0/testnet2
+last updated for testnet2
 
 ---
 
 ## Why Use Testnet?
 
-Testnet2 is a wonderful place where you can experiment with the Decred applications without worrying that a mistake will cost you real money. It is actually recommended that people use the Testnet2 to learn the basics of the Decred software or any new features.
+The testnet is a wonderful place where you can experiment with the Decred applications without worrying that a mistake will cost you real money. It is actually recommended that people use the testnet to learn the basics of the Decred software and any new features.
 
 Decred is currently on its 2nd Testnet, also known as testnet2. Testnets are periodically reset to help keep a manageable blockchain file size. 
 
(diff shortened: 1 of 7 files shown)


jcvernaleo:
Could you edit your comment to say Closes #78 instead of would close?  That way when we do merge it after release the issue actually gets closed as well.

docs/getting-started/using-testnet.md
```diff
@@ -0,0 +1,46 @@
+# Using Testnet Guide
+
+last updated for v1.0.0/testnet2
```
jcvernaleo: I like the mention of testnet2 but I don't think you need the v1.0.0 version number here.  That just happens to be when testnet2 comes out but it will still be around for a fair number of versions after that.

docs/getting-started/using-testnet.md
```diff
@@ -0,0 +1,46 @@
...
+
+---
+
+## Why Use Testnet?
+
+Testnet2 is a wonderful place where you can experiment with the Decred applications without worrying that a mistake will cost you real money. It is actually recommended that people use the Testnet2 to learn the basics of the Decred software or any new features.
```
jcvernaleo: Might be confusing to refer to it as testnet2 here.  Probably should just say testnet.  You could have a little blurb somewhere to say that the internal name is testnet2 since we reset the network periodically and this is the 2nd iteration of it, but generally I think we should just call it testnet.

jcvernaleo: As you can see I've been commenting as I go along and you do have that blurb right below :)

I still think we should generally call it testnet instead of testnet2.

docs/getting-started/using-testnet.md
```diff
@@ -0,0 +1,46 @@
...
+
+## How to Run a Testnet Node
+
+Running a testnet2 node is incredibly easy. You application of choice will need to download the testnet2 blockchain, and you will need to create a new wallet file for testnet2 use. Your mainnet blockchain and wallet files will remain untouched. Switching between the two is incredibly easy.
+
+---
+
+## Paymetheus
+
+To launch `Paymetheus` on testnet2, simply launch the `Decred Testnet` application instead of the usual `Decred` application. You will then be walked through the wallet creation (the same steps in the [Paymetheus Setup guide](/getting-started/user-guides/paymetheus.md)). After `dcrd` finishes syncing in the background, you'll then be able to fill your `Paymetheus` wallet with tDCR and familiarize yourself with the software.
```
jcvernaleo: Do we ever call the coins tDCR anywhere else?  Not saying I don't like the idea but I don't think I've ever seen it as anything other than testnet coins elsewhere.

docs/getting-started/using-testnet.md
```diff
@@ -0,0 +1,46 @@
...
+
+---
+
+## Command-Line Suite
+
+To launch `dcrd` and `dcrwallet` on testnet2, simply add the `--testnet` flag to your launch command. Alternatively, you could put `testnet=true` in all of your config files, but it's usually much faster to use the startup flag.
```
jcvernaleo: The sample files all use `testnet=1`.  Both work but I think we should be consistant.

docs/getting-started/using-testnet.md
```diff
@@ -0,0 +1,46 @@
...
+
+---
+
+## Acquiring Testnet2 Coins
+
+You can acquire coins through the [Decred Testnet Faucet](https://faucet.decred.org). Please return any coins to the address listed at the bottom of that page when you're done playing with testnet2.
+
+---
+
+<!-- TODO: ADD DECREDITON SECTION -->
```
jcvernaleo: FYI, decrediton (on master anyway) has a `--testnet` and `--mainnet` flag when you start from the command line.  The only catch is that if you use a flag, it will start in that mode next time (so you need to use the other flag to turn it back.

jcvernaleo:
Also, you should rebase this to keep it up to date for when it can go in.

jcvernaleo:
Just a reminder that if we want this ready for soon after the release might be a good time to both rebase and resolve the issues from review.

docs/getting-started/using-testnet.md
gratefulcheddar: Can't figure out how to do that on my Mac. Any ideas?

jcvernaleo: In a terminal:
```
jcv@boo ~ $ /Applications/decrediton.app/Contents/MacOS/decrediton --testnet
```
A bit awkward, but the other option is editing the config file `~/Library/Application\ Support/decrediton/config.json` and changing the line `"network": "testnet",`

jcvernaleo:
Changes in there look good.

jcvernaleo:
I think this is ready to go in.  @gratefulcheddar you still have the title as 'do not merge'.  If it is really ready, could you change the title?  If not, what's left?

gratefulcheddar:
I just need to rebase and add the new file to the translations I believe. Will try to take care of this later tonight.

jcvernaleo:
@gratefulcheddar looks like this needs a rebase again.  I think you did the previous rebase while I was out so sorry I missed it.  Once you rebase I'll merge it in right away this time.

gratefulcheddar:
@jcvernaleo rebased! Sorry that took so long

jcvernaleo:
Awesome, thanks!

PR #169, opened 2017-04-05
Add Spanish constitution

### LINKED ISSUE #18: drcdocs and the wiki
https://docs.decred.org/
https://wiki.decred.org/Main_Page

Why are those two separate and not one thing? What purpose each of them serves? When should I visit the one over the other?

jholdstock:
Closes #18


jcvernaleo:
Does this complete take care of #18, partially take care of it?  Just linking to the issue in the comment is confusing.

jholdstock:
Apologies - that was meant to be "Closes...". I've updated it now.

The other two Spanish pages on the wiki contain information which doesn't directly map onto anything in dcrdocs so I think this is the last of the info on there which is worth copying over.

jcvernaleo:
Awesome, glad to get this in!

Issue #170, opened 2017-04-06
Finally kill wiki


jcvernaleo:
Not 100% a docs issue but better for tracking here than anywhere else.

We need to find and change any places where we link to the wiki on sites we control and either remove them or replace them with links to the docs.  When this issue is closed, @dajohi can archive and kill the wiki.  Might want to redirect the subdomain domain to docs.decred.org at that point.

- [x] bitcointalk post
- [X] main webpage
- [x] docs pages (there are some links)
- [X] twitter? - doesn't link to wiki
- [x] forum (not in posts obviously)
- [x] /r/decred
- [x] Terms and rules on forum points to legacy wiki
- [x] package readmes
- [x] links to verifying sigs in release and binary repos

If anyone finds any other links, please comment here and I'll add them to this list.  If you take care of one of them, also comment so I can check it off.

jolan:
bitcointalk post is done.

jcvernaleo:
And looks like the twitter doesn't link to it at all.

jholdstock:
dcrdocs is covered by PR #171 above.

I think there is only one reddit post we need to look at and its here: https://www.reddit.com/r/decred/comments/3x8r5n/decred_an_overview/

I've taken a reasonably detailed look at the wiki and can't find anything which references the wiki unless I dig really far back into old posts. I don't think we have anything to worry about on the forum.

jcvernaleo:
Not really worried about reddit posts (since we will just has all links go to the main doc page).  I just wanted to make sure we didn't link to the wiki on the sidbar or something like that on the subreddit.

jcvernaleo:
I meant that last comment for both reddit and the forum (in case the main page of the forum or something linked to wiki, not posts).

jcvernaleo:
Confirmed the reddit had nothing on the sidebar

jcvernaleo:
Forum is fine too.

jholdstock:
We missed one! 

The `Terms and Rules` link at the bottom of the forum front page points to this wiki page:
 https://wiki.decred.org/Rules

The obvious solution is to move this page to dcrdocs and update the forum config so it points to the new location. We could also adapt these rules to make them cover slack comms as well as forum comms. I haven't given the rules an in-depth review so they may not even need any updating.

jcvernaleo:
I don't think I ever saw that page before.

[...]

PR #171, opened 2017-04-06
Remove references to wiki.decred.org

jholdstock:
Part of #170


jholdstock:
I included a tiny bugfix in this as well, just correcting a broken link. Hope thats ok.

jcvernaleo:
As long as the bugfix is in the commit message (and it is) I'm good with it going in with this.  For code I might feel different, but for docs, I can definitely live with it.

Issue #179, opened 2017-04-10
need to migrate mathjax to a different CDN


jolan:
https://www.mathjax.org/cdn-shutting-down/

gozart1:
I managed to get mathjax loading from same origin by adding the entire mathjax dependencies locally.

But I never made a PR for it b/c the only way I could manage to workaround the live reload issue referenced here https://github.com/decred/dcrdocs/issues/49 was to run the doc server via `mkdocs serve --no-livereload`

I can make a PR for this if folks think that's a reasonable compromise

jcvernaleo:
@gozart1 I've never really seen the livereload work very well anyway.

How much extra stuff do we need in the repo if we do add mathjax?  I saw some talk about how to slim things down in that like from @jolan.  I'd of course be happiest if we could make that monstrosity as slim as possible.

Also, do you know if there is a way to specify --no-livereload as the default (in the .yaml file maybe)?  That would make this solution even nicer.

That said, I'm in favor of no external js if we can.

jholdstock:
There's a solution for this which maintains the live reload functionality in the PR above. The PR includes a detailed description of the changes so I won't duplicate it here. I haven't spent long on this (I already had the first commit and the second is just the output of a 'sed' command) and I won't spend more time on it unless we decide its the approach we want to take as I know @gozart1 has a potential solution as well.

gozart1:
@jholdstock Nice! Your PR looks pretty much exactly what I had + the additional clever sed hack.

jcvernaleo:
This should have been closed with #185

PR #185, opened 2017-04-12
Host MathJax locally

jholdstock:
Relates to #179.

This PR adds the MathJax library so it can be hosted locally. It should not be merged yet because MathJax consists of >30k files at >70MB and we don't need all of this for our project. We can trim some of the fat before it's merged.

There are two commits here; the first simply adds MathJax unmodified to the dcrdocs repo and uses it within the site. This should work fine while hosting in production but causes issues when using `mkdocs serve` without the `-no-livereload` flag. This is because mkdocs uses Tornado server internally and Tornado server attempts to find the string `</head>` in *_any_* file and inserts it's live reload js before that point. You can see that happening [here](https://github.com/tornadoweb/tornado/blob/master/tornado/web.py#L759-L782) in the Tornado source. This causes issues because MathJax constructs HTML documents dynamically inside its .js files, and obviously when Tornado modifies these files it causes issues with MathJax functionality.

To work around this, I have created the second commit which replaces all instances of `</head>` with `</head >` in the MathJax .js files using:

`find tpldcrdocs/js/MathJax-2.7.0/ -type f -name "*.js" | xargs sed -i 's/<\/head>/<\/head >/g'`

The generated HTML is still valid and the javascript functionality remains unaffected because the only places where the modification has happened is inside string literals. I have tested this fairly extensively in Chrome and have come across no strange behaviour in either Tornado's live reload or MathJax - it seems that both are working perfectly.


jcvernaleo:
So after thinking about it, I'm good with this solution.

I'd like us to slim down the mathjax install a little before it goes in if possible though.

@jolan @dajohi do either of you object to doing it this way?

jolan:
Nope, fine with me too.

dajohi:
OK on hosting locally.  Will approve after slimmed down.

jholdstock:
OK so I've done a good bit of trimming, mostly using the official-but-deprecated MathJax guide [here](https://github.com/mathjax/MathJax-docs/wiki/Guide:-reducing-size-of-a-mathjax-installation/1814429ed1e97bfb7675c0fd400804baa9287249).

From the original 34k files at 74MB we are down to 1.2k files at 8MB. I did disable the context menu to achieve this but it was somewhat extraneous anyway. It's probably possible to remove a couple more MB but I couldn't do this and confidently say it's not affecting functionality. 

I am now removing the "(do not merge)" label from this PR.

jholdstock:
Oh just one more thing to note, all of the "trimming" commits are purely file deletions apart from a few lines which were added to `./tpldcrdocs/js/base.js`.

https://github.com/jholdstock/dcrdocs/blob/6520682f0d9c7f1c91962dde8b230f0f8ec2830e/tpldcrdocs/js/base.js#L51-L54

These two lines are needed to disable the context menu and to tell MathJax not to use SVG fonts.

gozart1:
image

jcvernaleo:
I'm all good with this now.  @dajohi if you are good with the size and approve we are good to go on it.

dajohi:
OK

jcvernaleo:
Great, going in.

PR #186, opened 2017-04-12
Standardise CLI param placeholders.

### LINKED ISSUE #182: Make placeholders for user input (passwords, etc) consistant
We are completely inconsistent in how we should that something in an example needs to have a user specified value.  I have seen (and done) things like:
```
dcrctl -u USER -p PASS getblockcount
```
```
dcrctl -u USER -p sekret getblockcount
```
```
dcrctl -u 'USER' -p 'PASS' getblockcount
```
```
dcrctl -u <USER> -p <PASS> getblockcount
```

We need to pick one and update all the docs to use it.  Suggestions are welcome.  I hate the use of quotes to indicated it (I've seen that trip up users too many times) so that is off the table.  I'd prefer `<USER>` myself but could be convinced of others.

jholdstock:
Closes #182

### DIFF
--- a/docs/advanced/program-options.md
+++ b/docs/advanced/program-options.md
@@ -6,7 +6,7 @@ The default options are sane for most users (although you may need to
 set your credentials. This means `dcrd` will
 work 'out of the box' for most users. However, there are also a wide
 variety of flags that can be used to control its behavior. The following section
-provides a usage overview which enumerates the flags.An interesting
+provides a usage overview which enumerates the flags. An interesting
 point to note is that the long form of all of these options (except
 `-C`) can be specified in a configuration file that is automatically
 parsed when dcrd starts up.
@@ -19,7 +19,7 @@ You can always get a list of all options by using the `-h` option.
 ## **Usage**
 
 ```
-dcrd [OPTIONS]
+dcrd <OPTIONS>
 ```
 
 ---
--- a/docs/advanced/program-options.md
+++ b/docs/advanced/program-options.md
@@ -102,7 +102,7 @@ the background. Let's have a look at some of the information
 available:
 
 ```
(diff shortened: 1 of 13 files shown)


docs/getting-started/startup-basics.md
```diff
@@ -53,14 +53,14 @@ At the very minimum, for `dcrd`, `dcrwallet`, and `dcrctl` to be able to communi
 4. Create an empty text file and save it as `dcrctl.conf` in `dcrctl`'s home directory.
 5. Choose an arbitrary username and password, these will only be used for each application to communicate via remote procedure call. The easiest configuration is to set them all equal.
 6. Inside `dcrd.conf`, underneath `[Application Options]`, add the following lines:<br /><br />
-        rpcuser=chosen-username<br />
-        rpcpass=chosen-password<br /><br />
+        rpcuser=<chosen-username><br />
```
gratefulcheddar: Put these lines in backquotes (`) otherwise Mkdocs tries to render the brackets as HTML tags.

jholdstock:
Done and I also found one more place to add the <> brackets.

docs/advanced/program-options.md
```diff
@@ -19,7 +19,7 @@ You can always get a list of all options by using the `-h` option.
 ## **Usage**
 
 ```
-dcrd [OPTIONS]
+dcrd <OPTIONS>
```
jcvernaleo: Actually, I think the square brackets were correct there.  This is not a single value that the user replaces with their own stuff but a whole bunch of stuff.  Much like the way most man pages do it:
```
SYNOPSIS
       ls [OPTION]... [FILE]...
```

jcvernaleo:
All good now.  Thanks!

PR #202, opened 2017-04-24
Update version numbers in docs to 1.0.0

jcvernaleo:


jrick:
Not sure we want these:

```
-This guide was last updated for v0.8.2
+This guide was last updated for v1.0.0
```

when the following doco wasn't touched at all.

jcvernaleo:
I went back and forth on that (and honestly I don't like this guide was last update for bit at all).  I decided that if the writing was accurate for v1.0.0 it should be bumped.  I could we convinced otherwise but what I've seen in having all sorts of different version numbers in there hurts more than it helps.

jrick:
a last updated date is more useful imo

tuxcanfly:
`git grep` found these, probably need to be updated

    de_docs/getting-started/install-guide.md:    For 32-bit computers, download the `decred_0.8.2-beta_x86.msi` file. <br />
    de_docs/getting-started/install-guide.md:    For 64-bit computers, download the `decred_0.8.2-beta_x64.msi` file.
    es_docs/getting-started/install-guide.md:    For 32-bit computers, download the `decred_0.8.2-beta_x86.msi` file. <br />
    es_docs/getting-started/install-guide.md:    For 64-bit computers, download the `decred_0.8.2-beta_x64.msi` file.

jcvernaleo:
@tuxcanfly thanks, no idea how I missed those since I was using grep also.  Fixed.

jcvernaleo:
@jrick I think I agree about the date (would help with the translations too) but I think that needs to happen in a seperate PR.  Would you be okay with this going in and me opening an issue to move to dates for things like that?

jrick:
sure

PR #208, opened 2017-04-27
Add new block header 'StakeVersion'.

### LINKED ISSUE #200: add stakeversion to block header specifications

jholdstock:
Closes #200


tpruvot:
its wrong, stakeversion was part of extradata

jholdstock:
You're right, i had missed the fact that ExtraData was reduced in length by 4 bytes when 4 bytes was allocated for StakeVersion. I will have to fix this in a couple of hours unless someone else gets to it first.

jholdstock:
I've just had a conversation with @davecgh and he has clarified that `StakeVersion` is a new field which comes after `ExtraData` and `ExtraData` has indeed been shortened.

I have also split `Nonce` out into its own field which comes before `ExtraData` with Dave's approval.

davecgh:
I would suggest using a non-zero value for the stake version.  Perhaps 3?  For example:

`0x03, 0x00, 0x00, 0x00`

tpruvot:
why did you merge the other one ? :/

jholdstock:
I've messed this branch up too much, I'm going to just create a new PR.

PR #212, opened 2017-04-28
Paymetheus - Voting address clarification

guyhalestorm:
Clarifies that Ds and Dc addresses are both valid targets for voting addresses.


guyhalestorm:
And because I was a moron and didn't put the gominer Linux guide in its own branch.... here it apparently is again.  *facepalm*

guyhalestorm:
Yay for newb mistakes.  I'll close this out and remake it.

jcvernaleo:
You don't actually need to close.  Make new branch with same name and force push `git push <REMOTE> <BRANCH> -f`

guyhalestorm:
Screw it, I can't make it work without it pulling in the old commits., I'll just make a new one.

PR #224, opened 2017-04-30
Add simnet guide to advanced section.

### LINKED ISSUE #188: Mainnet vs Simnet vs Testnet
The difference between these, what they should be used for. How to use them.

I think the only thing we already have on this topic is [the different type of addresses]( https://docs.decred.org/faq/wallets-and-seeds/#9-what-are-the-different-types-of-addresses)

[Here](https://gist.github.com/davecgh/2992ed85d41307e794f6) is a good resource explaining simnet for btcd, which is mostly analogous for dcrd.

Could mention testnet faucet here as well - the only place we currently mention it is in the Testnet voting guide, and I assume we will be replacing that soon in favour of a Mainnet voting guide

jholdstock:
Closes #188.

Info taken from https://gist.github.com/davecgh/2992ed85d41307e794f6#file-using_btcd_simnet-md

### DIFF
--- a/de_docs/advanced/simnet.md
+++ b/de_docs/advanced/simnet.md
@@ -1,0 +1,68 @@
+# **Simnet**
+
+---
+
+## **Overview**
+
+When developing Decred applications or testing potential changes, it is often extremely useful to have a test network where transactions are actually mined into blocks and difficulty levels are low enough to generate blocks as needed.
+
+In order to facilitate these scenarios, dcrd provides a simulation network (`--simnet`).  The following is an overview of the most important properties that distinguish it from the main network:
+
+* The difficulty starts extremely low to enable fast CPU mining of blocks
+* Networking changes:
+    * All code related to peer discovery and IP address dissemenation is disabled to help ensure the network remains private
+    * The peer and RPC network ports are different
+    * A unique network byte sequence is used in the peer-to-peer message protocol so the blocks can't accidentally be crossed with the main network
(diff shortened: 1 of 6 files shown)


de_docs/advanced/simnet.md
```diff
@@ -0,0 +1,68 @@
+# **Simnet**
+
+---
+
+## **Overview**
+
+When developing Decred applications or testing potential changes, it is often extremely useful to have a test network where transactions are actually mined into blocks and difficulty levels are low enough to generate blocks as needed.
```
jcvernaleo: Not sure what the part 'transactions are actually mined into blocks' even means.  Probably can just drop that bit.

de_docs/advanced/simnet.md
```diff
@@ -0,0 +1,68 @@
...
+
+    `$ dcrwallet --simnet --username=<username> --password=<password>`
+
+* Create a new simnet Decred address:
+
+    `$ dcrctl --simnet --wallet --rpcuser=<username> --rpcpass=<password> getnewaddress`
+
+* Stop the initial dcrd process and restart it with the mining address set to the output from the previous command:
+
+    `$ dcrd --simnet --rpcuser=<username> --rpcpass=<password> --miningaddr=S....`
```
jcvernaleo: Shouldn't the mining address also be in the form `<S....?>` ?

jholdstock:
Review comments addressed

jcvernaleo:
Looks all good now.

PR #301, opened 2017-05-17
Set build_docs.sh exit status.

### LINKED ISSUE #300: build_docs.sh output/return code
Two goals here:

- When a dev runs the script its not obvious at a glance whether it has succeeded or failed. This can be improved by:
    - Stopping the build when the first error is encountered
    - Outputting a clear success/failure message

- Get the script ready for CI
   - Currently the script returns exit status 0 regardless of what happened in the run. We need to return a non-zero code when an error occurs

jholdstock:
Closes #300

### DIFF
--- a/build_docs.sh
+++ b/build_docs.sh
@@ -65,10 +65,4 @@ for L in $LANG; do
 
 done
 
-echo
-echo "+-------------------------+"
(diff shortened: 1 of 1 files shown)


jcvernaleo:
Looks all good now!

PR #321, opened 2017-06-19
Update general.md

underscoreingsoc:
Added a question to the FAQ section on Decred's scalability and discussion from @davecgh on the topic.


raedah:
ACK

jcvernaleo:
There was a lot of discussion on slack about this and some people (myself included) thought we should be much more careful with the word 'infinitely'.  Also the two different meanings of scaleability in the answer seem to be a problem for docs.  I like the text davec has, but not for that question.  Maybe it could be split out for what people normally mean with scaleability and what davec means as 'adaptability' or something like that.

Either way I think this needs work (as the discussion that wasn't even started by me on slack indicated) before it can go in.

jcvernaleo:
Discussed with @underscoreingsoc and he said to close it rather than fix this one up.

PR #336, opened 2017-07-28
Update the web wallet doc about mainnet and testnet

teknico:
The web wallet doc states that it only works on testnet. It actually works on mainnet too, and it's even the default, as set in https://github.com/decred/copay/issues/52 . This change updates the doc to reflect the current situation.

(I was initially thrown off by that paragraph but decided to try anyway, and it did work on mainnet, so here it is. :-) )

### DIFF
--- a/docs/getting-started/user-guides/web.md
+++ b/docs/getting-started/user-guides/web.md
@@ -49,7 +49,7 @@ assume you are just starting out so click `Get Started`.
 > Step Two
 
 A wallet will be generated for you.  Under `Personal Wallet` on the
-left, choose `Mainnet` (the defaul) if you want the wallet to hold
+left, choose `Mainnet` (the default) if you want the wallet to hold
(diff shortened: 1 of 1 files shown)


docs/getting-started/user-guides/web.md
```diff
@@ -48,13 +48,13 @@ assume you are just starting out so click `Get Started`.
 
 > Step Two
 
-A wallet will be generated for you and you will see this screen:
+A wallet will be generated for you.  Under `Personal Wallet` on the
+left, choose `Mainnet` (the defaul) if you want the wallet to hold
```
sndurkin: Minor typo, should be "(the default)".

teknico: Thank you for the catch.

teknico:
The interface changed and the whole section is obsolete, closing.

Issue #350, opened 2017-08-29
macOS Inconsistency 


aldoiljazi:
**Mac** and **OS X** should be changed to **macOS**, since Apple's operating system name has changed. [https://www.apple.com/macos/sierra/](https://www.apple.com/macos/sierra/)

jolan:
Closed via #351.

BTW, you need to put the "fixes ..." in the first comment.

PR #351, opened 2017-08-29
macOS Inconsistency

aldoiljazi:
Changed **Mac** and **OS X** to **macOS**, since Apple's operating system name has changed. [https://www.apple.com/macos/sierra/](https://www.apple.com/macos/sierra/)


aldoiljazi:
Fixes #350

PR #362, opened 2017-09-03
Translate to Chinese.

AbnerZheng:
I have created Chinese translation directory, and translate some files. And if decred is willing for translating to Chinese, I will continue this work.


peterzen:
This should be imported into Transifex instead of be merged (the old translated files are being deprecated in favor of TF)

AbnerZheng:
Sure, I've been add as a translator of Decred project.  I'll merge the content to Transifex. So I close these two MR.

PR #396, opened 2017-10-16
Updated list of exchanges

nickboariu:
Included litebit.eu, cryptopia and exodus


fai04d:
Do we maybe want to swap the position of Bittrex and Poloniex given the trouble they've had recently? People generally just go to the first link they see so we should be careful about what's at the top even though we don't promote any exchange other another.

PR #410, opened 2017-11-30
material-theme and dependencies

denyszayets:
This is only initial batch of new files.


denyszayets:
same files are here https://dcrdocs.netlify.com/

PR #465, opened 2018-03-01
Update obtaining-dcr.md

jzbz:
Cleaned up language and style.
Sorted the list so highest volume exchanges are at the top.
Linked directly to exchange DCR listings.
Removed Yobit because DCR withdrawals do not work.
Removed Changelly as they no longer support DCR.
Removed Exodus as it's not an exchange and uses ShapeShift.

### DIFF
--- a/docs/getting-started/obtaining-dcr.md
+++ b/docs/getting-started/obtaining-dcr.md
@@ -4,15 +4,14 @@
 
 ## <img class="dcr-icon" src="/img/dcr-icons/Info.svg" /> Overview
 
-This article discusses some of the ways a user can acquire Decred
-coins.
+This article discusses some of the ways a user can acquire Decred.
 
 The five basic ways are:
 
-* Purchase Decred for BTC through an exchange
-* PoW mine for Decred
-* PoS mine for Decred (this requires you to already have some Decred)
-* Contribute to Decred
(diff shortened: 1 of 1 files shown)


docs/getting-started/obtaining-dcr.md
```diff
@@ -21,56 +20,53 @@ The five basic ways are:
...
+### Instant Exchanges
+
+Instant exchanges allow users to convert from one cryptocurrency to another without the need for registration. Do note that higher fees usually apply, i.e., the price will include a fee for the convenience. You should compare prices among all the exchanges below.
+
+* [ShapeShift](https://shapeshift.io/)
+* [Flyp](https://flyp.me/)
+* [Godex](https://godex.io/)
+* [CoinSwitch](https://www.coinswitch.co/)
+* [Evercoin](https://evercoin.com/)
+* [BitPlace](https://www.bitplace.exchange/)
```
alexlyp: We should probably remove this right?

jzbz: BitPlace is dead, I think the others are still fine unless someone reported an issue.

alexlyp: Right just bitplace to remove

Issue #471, opened 2018-03-13
Broken icons


jholdstock:
Visit a page other than the home page (eg https://docs.decred.org/getting-started/beginner-guide/) and the icons are broken in the navigation menu.

404 on `font-awesome.min.css` and `material-icons.css`

obxium:
🎉 It looks like this is now fixed!

PR #474, opened 2018-04-03
Update theme override paths for fonts & icons

obxium:
- Absolute path for Font Awesome style
- Include base url in path for Material Icons style

This should help with pages such as:

https://docs.decred.org/getting-started/using-the-block-explorer/

which currently 404 on the Font Awesome and Material CSS in the live documentation site.


obxium:
This could resolve #471 as well

Issue #240, opened 2017-05-02
Glossary


jcvernaleo:

Arriu:
I've taken on this issue. I'm going to tackle it in two stages. First i'm building a list by going over https://docs.decred.org/ from top to bottom selecting any words I think might be hard to understand to people new to the space. I'm being quite extensive, since i think redundancy can't hurt here. Anyone is free to add to the list if they feel there are words missing. Second I'll be writing actual Glossary itself, making as much use of citable sources as possible.

dajohi:
@arriu can you submit a PR?

Arriu:
I'm sorry I hadn't communicated this here earlier. I haven't found any time to spend on DCR in general in the past half year and I can't forsee when i'll be able to do so again just yet. Please remove me from this issue. I'd alrdy communicated on slack that if anyone else showed up they were absolutely free to take this up. I also shared the incomplete list of terms I'd made there. Unfortunately, I've been away from it all for so long I don't even remember how to use git and github. I hope this hasn't caused anyone inconvenience and I still hope to contribute in the future.

jholdstock:
Closing this - glossary work will be tracked in #589

jholdstock:
Closing this - glossary work will be tracked in #589

jholdstock:
Closing this, will use #589 for glossary work

PR #500, opened 2018-05-25
Glossary list

Arriu:
This list is a first step to resolving issue #240. i'm building a list by going over https://docs.decred.org/ from top to bottom selecting any words I think might be hard to understand to people new to the space. I'm being quite extensive, since i think redundancy can't hurt here. Feel free to add to the list if you think there are words missing. When I've covered the entire docs I'll start working on the actual Glossary, making as much use of citable sources as possible.


xaur:
Nice list!

Off top of my head I would add these: consensus vote, emission, inflation, dcrd, dcrwallet, dcrctl, decrediton, immature (I see mature is there), revoked/missed/expired tickets.

Also, please update your PR to only include the glossary file, the diff currently shows modifications of some other files.

PR #520, opened 2018-05-30
fix generated js and css filenames.

dajohi:


jholdstock:
This can be closed now @dajohi

PR #532, opened 2018-06-10
 Update inflation.md

hypernoob:
- Updated the supply schedule
- Updated info about max supply
- Added info about reduced reward before block 4096


hypernoob:
I used the TestBlockSubsidy function in dcrd to verify total supply and when the last block reward will be generated ([patch](https://hastebin.com/raw/opofenetew)).

[Results](https://hastebin.com/raw/ejafetuyib)

The last block reward will be produced at height 11003903 (~38208 days since genesis i.e. 18 September 2120).

davecgh:
Yeah, I rounded up to 2121 because it's so far out that the exact month is unlikely end up being the case.  It will be close, but it all depends on the ultimate block production rate.

Also, there will very likely end up being a vote for a one-time tail emission that creates the coins never produced due to missed votes and fees lost through invalidation which I figured would extended it into 2121.

PR #536, opened 2018-07-02
Update using-testnet.md

RichardRed0x:
updating Decrediton testnet instructions

### DIFF
--- a/docs/getting-started/using-testnet.md
+++ b/docs/getting-started/using-testnet.md
@@ -21,6 +21,7 @@ Running a testnet2 node is incredibly easy. You application of choice will need 
 ## Decrediton 
 
 You can switch Decrediton between mainnet and testnet by editing the config.json file and setting network: testnet. Config.json can be located in the following folders:
+
 | OS      | Decrediton data directory                           |
 | -------:|:---------------------------------------------------:|
 | Windows | `C:\Users\<your-username>\AppData\Local\Decrediton` |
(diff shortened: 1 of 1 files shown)


jholdstock:
Hi Richard,

In all other places when we describe OS specific paths, we also include macOS. Could you add that please.

Could you also wrap the paths in backticks (`) rather than quotation marks (").

It might be nice actually to list the paths in a table, for easier readability. An example of this can be seen here:

https://docs.decred.org/getting-started/user-guides/decrediton-setup/

RichardRed0x:
Copied that table over, I agree it's nicer.

docs/getting-started/using-testnet.md
```diff
@@ -20,7 +20,13 @@ Running a testnet2 node is incredibly easy. You application of choice will need
 
 ## Decrediton 
 
-You can switch Decrediton between mainnet and testnet by changing the "Network" setting in the Settings tab, and then pressing the save button. You will have to close Decrediton and open it again for the change to take effect.
+You can switch Decrediton between mainnet and testnet by editing the config.json file and setting network: testnet. Config.json can be located in the following folders:
+
+| OS      | Decrediton data directory                           |
+| -------:|:---------------------------------------------------:|
+| Windows | `C:\Users\<your-username>\AppData\Local\Decrediton` |
```
hypernoob: I suggest changing this to ```| Windows | `%localappdata%\Decrediton` |```

RichardRed0x:
Done

PR #541, opened 2018-07-07
Replace links to (mainnet.decred.org) with (explorer.dcrdata.org)

### LINKED ISSUE #535: Replace links to (mainnet.decred.org) with (explorer.dcrdata.org)
As per @jholdstock's comment 

> In agenda-voting.md we have

`The easiest method to track your how your tickets actually voted is to use the [block explorer](https://mainnet.decred.org) to view each of your votes.`

> In dcrd-setup.md we have the following sentence:

`... the last block processed against the last block height on [the official block explorer](https://mainnet.decred.org/). `

There might be more of these links in the docs, ideally we replace all of them with a link to dcrdata.

hypernoob:
Close #535


hypernoob:
Blocked until decred/dcrdata#338 is resolved.

jholdstock:
Created #549 to address the broken links

chappjc:
@jholdstock There may be no fix to this on dcrdata's end.  The ability of an explorer to show sidechain blocks depends entirely on whether or not dcrd has every seen them.  A freshly deployed node will have no knowledge of any side chains.  A long running dcrd will, however.  We can identify a side chain block that dcrdata/dcrd does know about on a particular node, but there's no guarantee it will not be lost in the future (e.g. upgrade, format machine, etc).

chappjc:
The best fix possible is implemented in https://github.com/decred/dcrdata/pull/717.  With that, any side chain blocks may be loaded in dcrdata.

However, as I mentioned, it will not be possible to find a side chain block if dcrd was not up and running when that fork happened.  The fork must be observed by dcrd.  A fresh sync of dcrd will lose all record of side chain blocks previously recorded.

PR #586, opened 2018-08-21
Note about forthcoming constitutional reform (#576)

### LINKED ISSUE #576: Constitution: add notice about incomplete systems
Following recent [discussion](https://www.reddit.com/r/decred/comments/96j9ai/transparency_in_payments_made_with_the_community/e43btbt/) on Reddit I came to a realization that some people (and trolls) might mis-read the constitution as something that should have been fully in effect since day one. Obviously this is not possible because "Decred was not built in a day".

To prevent any confusion I suggest adding a notice on top of the page:

> Note: The constitution was written with the future in mind and has various portions that are contingent upon systems that are not yet complete.

While amendments to the constitution must happen with stakeholder approval, a notice on top is technically "outside" the document.

RichardRed0x:
Some aspects of the constitution are obsolete, some are yet to be put into practice, Politeia proposals will soon bring it up to date. Closes #576

### DIFF
--- a/docs/getting-started/constitution.md
+++ b/docs/getting-started/constitution.md
@@ -1,8 +1,8 @@
 # <img class="dcr-icon" src="/img/dcr-icons/Constitution.svg" /> Decred Constitution
 
-> Note: The constitution was written prior to Decred's launch, some portions of it are no longer relevant and some are yet to be realized. Politeia will be used, when it launches, to propose amendments to the constitution for ratification by Voters.
+!!! Note
+The constitution was written prior to Decred's launch, some portions of it are no longer relevant and some are yet to be realized. Politeia will be used, when it launches, to propose amendments to the constitution for ratification by Voters.
 
----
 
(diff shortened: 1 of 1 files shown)


xaur:
I like the text, another improvement from last version posted in #576.

Regarding style, quote (`> `) will do, but there is some "admonition" styled block that could be used here. [Example](https://docs.decred.org/getting-started/user-guides/decrediton-setup/). @jholdstock could comment on how to use it.

jholdstock:
I would say we can also remove the horizontal rule after the info panel. It looks a bit out of place having both.

The horizontal rule is inserted by the line `---`

RichardRed0x:
Thanks, I didn't know about automatically closing issues.
I updated to use the info box.

xaur:
Will it work with capitalized Note?

xaur:
Also the 4-space indent may be necessary.

P.S. Idealy there would be a test deployment somewhere to see and discuss deployed changes.

RichardRed0x:
I think Note is just a title, could be anything.
I don't know about the spacing, can't preview it. Happy to go either way on that, @jholdstock's call.

jholdstock:
- The four spaces are totally necessary - this is how the admonition plugin knows which text should be included in the admonition panel.
- "note" is actually the type of the admonition - the fact that the title is also "Note" is coincidental. We should keep it lower case as per the guidelines (https://squidfunk.github.io/mkdocs-material/extensions/admonition/#note) and also to be consistent with our usage of admonitions elsewhere in the docs.
- Agree with @s-ben - semi-colon would be more correct than a comma.

RichardRed0x:
All fixed.

PR #601, opened 2018-09-04
Update versions of mkdocs and material theme.

### LINKED ISSUE #596: Investigate updating mkdocs and mkdocs-material theme
mkdocs has had a major version update - we are running 0.17.3 and version 1.0.3 has been released. 

There is also a new version of our theme, material. We are using 2.7.1, and the latest is 3.0.3.

We should look into feasibility of upgrading to the latest releases, the effort required, and the benefits of doing so.

jholdstock:
- No longer need to override any of the material theme.
- Fix a path issue in inflation.md - this is highlighted by the new version of mkdocs.
- Update Docker image to use python 3.7.

Closes #596


jholdstock:
This has a conflict, and also I need to upgrade the README before this is merged.

Issue #608, opened 2018-09-11
Updating the Blockchain parameters


dmigwi:
After my research I realised that the maximum size of block allowed in decred is **1MB** while the maximum size of a transaction allowed is **393,216 bytes**.

Currently what is available of the documentation in https://docs.decred.org/advanced/blockchain-parameters/ says otherwise..
<img width="571" alt="screen shot 2018-09-11 at 06 47 55" src="https://user-images.githubusercontent.com/22055953/45337105-1c076e80-b58f-11e8-85ca-0575452d06a7.png">

dmigwi:
@davecgh 👆 Is the above accurate?

davecgh:
The documentation is correct.  As @hypernoob linked to, the maximum size for a block on `mainnet` is indeed 393216.  [1000000 bytes](https://github.com/decred/dcrd/blob/566bb6625b83738cb97ab826d5d77e862d44e92c/wire/msgblock.go#L28) is maximum size the message on the wire is allowed to be.

PR #609, opened 2018-09-11
Update max block

s-ben:
Updates max block size to 1M (if true) (#608 )


hypernoob:
Please close per https://github.com/decred/dcrdocs/issues/608#issuecomment-420283036

s-ben:
Called out as untrue, so no need to fix.

Issue #546, opened 2018-07-14
Link to dcrdata for blockchain size


jholdstock:
In the decrediton user guide, we advise that downloading the whole blockchain is required to run decrediton. 

From the page https://docs.decred.org/getting-started/user-guides/decrediton-setup/

> At the end of May 2018 the Decred blockchain was 3.3GB.

This PR for dcrdata adds a chart showing the current blockchain size: https://github.com/decred/dcrdata/pull/463

Once this change is live on https://dcrdata.org, we can link to it from the docs to provide a more accurate and up-to-date figure

noahpierau:
Started working on this issue, expect a PR soon!

noahpierau:
According to https://alpha.dcrdata.org/charts the blockchain size today is <2.4G

Did the changes to the blockchain database have this much effect on the file size?

noahpierau:
Preliminary changes: https://github.com/noahpierau/dcrdocs/blob/master/docs/getting-started/user-guides/decrediton-setup.md

PR #614, opened 2018-09-14
Link to dcrdata for blockchain size

noahpierau:
This PR closes issue #546


jholdstock:
I'm happy with this changes, but it would be great to understand the discrepancy between 3.3GB and 2.4GB blockchain size before we merge it.

The 3.3GB figure came from running decrediton on a local machine and checking the size of the blockchain db stored on disk. It would be interesting to understand how dcrdata calculates the chain size, especially the historic size.

papacarp:
The size as reported on dcrdata is a sum of all the individual block sizes.    As an example, this block is 6.2kB (https://explorer.dcrdata.org/block/275632).  We just sum all entries in our database for block size and report the result.  Which is approximately 2.6G (2,674,813,446).  Our database has recently started collecting blocks that are not valid and not part of the main chain as well and as of now those are also included in our totals.  The sum of all invalid and non-mainchain blocks is only 39KB however, so not a big deal.

chappjc:
To reiterate @papacarp, the size dcrdata reports is the sum of the sizes of all serialized blocks, as opposed to the disk space occupied by the database.  There is quite a bit of metadata and other DB overhead that adds to the size on disk.

noahpierau:
Based on the feedback I changed the final sentences. I also added the notion of SPV, since it doesn't require a full blockchain download. This will make the information more future-proof.

docs/getting-started/user-guides/decrediton-setup.md
```diff
@@ -8,7 +8,7 @@ Last updated for v1.2.1.
 
 !!! info "Blockchain Download"
 
-	Decrediton needs to download a full copy of the Decred blockchain before it can be used. This means that a large download must be completed and a large amount of storage space must be be available on the device where Decrediton is intalled. You can check the current size of the Decred blockchain via [dcrdata charts](https://explorer.dcrdata.org/charts#blockchain-size). Total required storage space for Decrediton may vary, depending on your device.
+	Without SPV enabled, Decrediton needs to download a full copy of the Decred blockchain before it can be used. This means that a large download must be completed and a large amount of storage space must be be available on the device where Decrediton is intalled. You can check the current size of all serialized blocks via [dcrdata charts](https://explorer.dcrdata.org/charts#blockchain-size). The total size of Decrediton will be bigger, since additional wallet data is stored on your device.
```
jholdstock: Typo: intalled -> installed.

noahpierau:
Fixed. Not sure how I missed that :smile:

PR #616, opened 2018-09-18
Various cleanups

dajohi:
- Use markdown comment-style
- remove stray </i>
- autocomplete is not valid for checkbox

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -2,14 +2,12 @@
 
 ---
 
-## <img class="dcr-icon" src="/img/dcr-icons/decred-logo.svg" /></i> What is Decred?
+## <img class="dcr-icon" alt="Decred logo" src="/img/dcr-icons/decred-logo.svg" /> What is Decred?
 Decred is a blockchain-based cryptocurrency with a strong focus on community input, open governance, and sustainable funding for development. It utilizes a hybrid Proof-of-Work (PoW) and Proof-of-Stake (PoS) mining system to ensure that a small group cannot dominate the flow of transactions or make changes to Decred without the input of the community. A unit of the currency is called a decred (DCR).
 
 ---
 
-<more information to link to Decred governance, tech and dev pages>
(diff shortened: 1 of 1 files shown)


theme_override/base.html
```diff
@@ -95,8 +95,8 @@
         {% endif %}
       </defs>
     </svg>
-    <input class="md-toggle" data-md-toggle="drawer" type="checkbox" id="drawer" autocomplete="off">
```
jholdstock: This file is a copy of the one provided by our mkdocs theme, hence why it is in the "theme_override" directory. We only made tiny changes to this file which were mostly for #470 and #598. These changes have now been included in the theme by default, so we will actually be removing the file soon, its already deleted in #601 

I therefore suggest we simply revert this change for now.

docs/index.md
```diff
@@ -2,24 +2,24 @@
 
 ---
 
-## <img class="dcr-icon" src="/img/dcr-icons/decred-logo.svg" /></i> What is Decred?
+## <img class="dcr-icon" alt="logo" src="/img/dcr-icons/decred-logo.svg" /> What is Decred?
 Decred is a cryptocurrency, similar to Bitcoin, with a strong focus on community input, open governance and sustainable funding and development. It utilizes a hybrid "proof-of-work" and "proof-of-stake" mining system to ensure that a small group cannot dominate the flow of transactions or make changes to Decred without the input of the community. A unit of currency is called a 'decred' (DCR). To ensure the integrity of the currency and prevent people from making fraudulent transactions or creating their own coins, Decred uses a method of recording transactions known as a blockchain.
 
 ---
 
-<more information to link to Decred governance, tech and dev pages>
```
jholdstock: To be honest I think we should just remove this comment and the comment below - it doesn't seem like whoever wrote them is going to come back and implement them.

dajohi:
updated

PR #631, opened 2018-10-09
cull remark-cli from package.json

dajohi:


jholdstock:
@peterzen could you take a look at this?

PR #590, opened 2018-08-29
[WIP] Jargon changes from  from “PoS Mining” to “(PoS) Voting”, and from “stakepool” to “Voting Service Provider (VSP)”

RichardRed0x:
This relates to Issue #581 – change of terminology from “PoS Mining” to “(PoS) Voting”, and from “stakepool” to “Voting Service Provider (VSP)”.
My intention is to submit this Pull Request as part of a Politeia proposal to gauge support for a change in terminology. While this may seem trivial, terms like “stakepool” are woven quite deeply into a number of platforms. Changing them all at once, to avoid confusion, is quite an undertaking, and therefore it seems appropriate that stakeholder support for the change should be established first.

So, **don't merge this PR yet.**

I identified the following pages where the old terms are used, there are probably more, if you know any please add them on this PR.
* getting-started/user-guides/how-to-vote/ - stakepool
* mining/proof-of-stake/ 
* mining/how-to-stake/ - and menu, how to vote
* faq/proof-of-stake/general/
* faq/proof-of-stake/solo-mining/
* faq/proof-of-stake/stake-pools/

Part of the way through editing the FAQ docs, I realized the information they contain is largely redundant with the more visible docs, so it probably makes sense to just leave them alone and add a note about older terms being used.

I have not changed filenames or the menu titles, yet. I can either do that as part of this PR or open a new one for it after this one is merged. It would be easier to wait until the titles within the documents are nailed down and update menus as a batch job.

PR #675, opened 2018-11-02
Adds glossary

### LINKED ISSUE #589: Request: Glossary page
Recommend creation of a glossary page that consists of short summaries of all the critical crypto and decred specific terms. When possible, each term can also link to specific docs entries for further read. 

A glossary page can essentially be used for a quick point of reference, as well builds a bit more consistency and be of help with localisation work. 

Few references:
https://crushcrypto.com/glossary/
https://www.upfolio.com/glossary

s-ben:
Adds glossary to docs ( Closes #589 ).

A draft of this was created in [dcrwords](https://github.com/RichardRed0x/dcrwords). Have ported this over to dcrdocs, polishing text and making terms into h4s (####). This allows us to link within the document, and also allows linking to specific terms from elsewhere in the docs. 

@RichardRed0x , I ended up just going with alphabetical. It wasn't clear how to categorize all the terms, and the TOC on the right panel makes it pretty easily searchable. Can re-organize if you have categories. 

The glossary didn't seem to fit any existing folders, so I've just put is as a separate file at the top level. Open to suggestions for another placement.

All the definitions seem solid to me, though we'll want a thorough read-through by someone that can catch any inaccuracies before we publish.


RichardRed0x:
Cool, I still think non-alphabetical would be better (I liked the rough unmarked groupings in the other repo) but do not feel strongly about it.

There was quite a bit of discussion in [this PR](https://github.com/RichardRed0x/dcrwords/pull/1) on the other repo if anyone is looking for more context around why this glossary is the way it is.

noahpierau:
I strongly prefer alphabetical, it's much easier to find what you are looking for. 

The glossary will be a great help for all the translators!

xaur:
Alphabetical is better because most glossaries I've seen were alphabetical and it is intuitive how to use them.

Grouping terms is also useful, but can be achieved with an index, which link to individual terms.

docs/glossary.md
```diff
@@ -0,0 +1,265 @@
...
+
+In each block, five tickets are called to vote. In addition to votes on any open consensus rule change proposals, each ticket votes to approve or reject the regular transaction tree of the previous block. If a majority of voting tickets vote No, the regular transaction tree of the previous block is rejected and those transactions are returned to the mempool.
+
+#### Command Line Interface (CLI)
+
+An interface a user interacts with via the command line - often referring to a CLI wallet, which is operated using various tools and commands.
+
+#### Confirmations
+
+The number of blocks that have been mined on top of the block when a transaction was made; more confirmations gives greater finality or confidence that the transaction will remain in the Decred blockchain.
```
jrick: Not quite accurate.  The most recent block is said to have 1 confirmation even though there are no blocks following it (the transactions within have one block confirmation).

docs/glossary.md
```diff
@@ -0,0 +1,265 @@
...
+
+A mechanism for obtaining free (testnet) tokens.
+
+#### Hash
+
+The output of a cryptographic hashing function that produces a fixed-size and unique hash value from variable-size input. 
+
+#### Hash function
+
+A cryptographic function that produces a fixed-size and unique hash value from variable-size transaction input. Decred uses the [BLAKE-256](https://docs.decred.org/research/blake-256-hash-function/) hashing function.
```
jrick: Many different hash functions are used by the software.  To make this clearer, this should say that the proof-of-work function is BLAKE256.

docs/glossary.md
```diff
@@ -0,0 +1,265 @@
...
+
+Miners help secure the network via Proof-of-Work (PoW) mining. Miners recieve block rewards and transaction fees for successfully validating blocks. 
+
+#### Mining Pool
+
+A group of miners who share (pool) their computational resources to mine DCR. When a member of a mining pool successfully mines a block, the rewards are split among all pool members in proportion to the hashpower they contribute to the pool. 
+
+#### Missed ticket
+
+Tickets that have been called but did not respond - these can be revoked, but do not grant a reward.
```
jrick: A ticket can also be missed even when a vote is created and published if a miner does not include it in the following block.

docs/glossary.md
```diff
@@ -0,0 +1,265 @@
...
+
+#### Orphan
+
+A valid block which is not built upon and therefore not included in the definitive blockchain.
```
jrick: There are two different things that "orphan" can refer to.  This is term for orphaned blocks, but it is also possible to have orphaned transactions (transactions with missing inputs -- their "parent" transactions are unknown making them "orphans").

docs/glossary.md
```diff
@@ -0,0 +1,265 @@
...
+
+#### Politeia
+
+A system for facilitating the submission and discussion of [proposals](https://proposals.decred.org/) in an environment with transparent censorship. 
+
+#### Private key
+
+A string of data that allows you to access and spend DCR stored in address associated with the private key. When sending DCR from one address to another, the wallet must use the private key of the sending address to sign the transaction.
```
jrick: Don't call this a string.  It's just binary data (actually, it's a number, but it's encoded as binary).

DCR can not be sent "from" an address.  Addresses are only capable of receiving coins, not sending.

docs/glossary.md
```diff
@@ -0,0 +1,265 @@
...
+
+Accounts allow you to keep separate records of your DCR funds. Transferring DCR across accounts will create a transaction on the blockchain.
+
+#### Agenda voting
+
+See [consensus rules voting](#consensus-rules-voting).
+
+#### Address
+
+The cryptographic hash of a given public key. Addresses are publicly viewable on the blockchain and can be used to send and receive transactions, similar to an email address.
```
jrick: Addresses can not send, only receive.

docs/glossary.md
```diff
@@ -0,0 +1,265 @@
...
+
+A system for facilitating the submission and discussion of [proposals](https://proposals.decred.org/) in an environment with transparent censorship. 
+
+#### Private key
+
+A string of data that allows you to access and spend DCR stored in address associated with the private key. When sending DCR from one address to another, the wallet must use the private key of the sending address to sign the transaction.
+
+#### Private passphrase
+
+A passphrate that is used to encrypt wallet.db files. wallet.db files encrypted with a private passphrase cannot be used without that passphrase.
```
jrick: The entire file isn't encrypted, but some data within the database is (most notably, private keys).

docs/glossary.md
```diff
@@ -0,0 +1,265 @@
...
+
+The mechanism by which ticket holders vote to approve blocks confirmed by PoW miners (thus providing a check on PoW miners), earn staking rewards, and vote on consensus rule changes and Politeia proposals. 
+
+#### Protocol rules
+
+See [consensus rules](#consensus-rules). 
+
+#### Public key
+
+A long string of alphanumeric characters generated algorithmically from a private key. The user’s public key is used to prove that a transaction was signed using their private key.
```
jrick: Also not a string (see comment about private keys).

docs/glossary.md
```diff
@@ -0,0 +1,265 @@
...
+#### Quorum
+
+The minimum level of participation required in order for a decision-making process to produce a valid outcome. Changes to the consensus rules require at least 10% of votes to be for or against the change in order to be valid. Politeaia proposals require 20%.
+
+#### Regular transaction tree
+
+The normal [transactions](https://www.reddit.com/r/decred/comments/66j4l4/decred_proof_of_stake_explained/dgjsyxd) in a block: sending DCR to an address, coinbase transactions such as PoW Miner reward and Treasury stipend.
+
+#### Reorg
+A reorganization of the blockchain in which a set of blocks are replaced by another (longer) set. The number of blocks replaced is the depth of the reorg.
```
jrick: The reorganize to the best chain may not necessarily follow the longest chain.  The best chain is the one with the most work.

docs/glossary.md
```diff
@@ -0,0 +1,265 @@
...
+
+An interval of 8064 blocks (~4 weeks) in which ticket holders can vote on consensus rule changes. A Rule Change Interval (RCI) consists of four Stake Version Intervals (SVI), which are 2016 blocks (~1 week). Once the conditions for a vote have been met during an SVI, voting is scheduled to begin on the first block of the next RCI. Because there are four SVIs per RCI, it can take up to 6048 blocks [3 SVIs] for the next RCI to begin.
+
+#### Rule change proposal
+
+A proposal to change the consensus rules of the Decred blockchain. Rule change proposals must be implemented in latent code within the software running the network's nodes. If the proposal passes, the latent code activates one month later. 
+
+#### Seed
+
+A series of 33 words that is used to generate the private key for a wallet. Anyone who has the seed for a wallet can spend any DCR it holds.
```
jrick: The seed is 32 bytes of data.  The 33 word encoding is just that, an encoding.

The wallet manages many private keys, not just one.

justinsantoro:
Idea:

You could use this formatting to generate the table of contents to allow users to jump to a specific letter:

```
# Main title

---

## A

…(all A words here)

---

## B

…(all B words here)

---
````

**Comment by RichardRed0x** [comment]
From [#writers channel](https://matrix.to/#/!lbzTjhzNbIaDbuAxkS:decred.org/$15411729057522NXffC:decred.org): Add an entry for Fully-validating node as the alternative to SPV

File: docs/glossary.md (hunk 9 above)
Comment (zubairzia0) [reply docs/glossary.md]: Would it be more intuitively understood if we say fastest chain? I say this because in decred a minority fork can have more hashpower yet be reorganized to the one with the majority of tickets.  Unless work encompasses hashpower and stake....

File: docs/glossary.md (hunk 1 above)
Comment (s-ben) [reply docs/glossary.md]: I googled around for how other glossaries defined this. All define 'Confirmation' (singular). I think this is better. Have taken a stab at a new definition that also includes the meaning of confirmations (plural). 

#### Confirmation

A transaction is confirmed when it has been included in a block that has been verified by the network (added to the blockchain). Once a transaction has been confirmed, it cannot be reversed or double spent. Each additional block added to the blockchain reconfirms all transactions in all previous blocks. The number of times a transaction has been confirmed (i.e. the number of blocks that include the transaction) is used as a measure of confidence that the transaction will remain in the blockchain. For example. a wallet or other service may requre a certain number of confirmations before it considers a transaction to be valid.

File: docs/glossary.md [hunk 11]
```diff
@@ -0,0 +1,265 @@
...
+
+A cryptographic function that produces a fixed-size and unique hash value from variable-size transaction input. Decred uses the [BLAKE-256](https://docs.decred.org/research/blake-256-hash-function/) hashing function. 
+
+#### Hashrate
+
+The number of hashes per second computed by miners on the network.
+
+#### Hybrid PoW/PoS
+
+In Decred, a [hybrid PoW/PoS system](https://docs.decred.org/research/hybrid-design/) is used, whereby blocks mined by PoW miners must be approved by ticket holders, which are pseudorandomly selected from the ticket pool. This provides a check on PoW miners and increases the overall cost of attacking the network. When a block is approved, 60% of the block reward goes to the PoW miner, 30% goes to the holders of  tickets called to vote (approve blocks), and 10% goes to the Decred Treasury to fund the project's development.
```
davecgh: Instead of saying "the project's development" I would just suggest "the project".  While I realize that all facets can be considered a form of "development", most people interpret that to mean writing code, and the Treasury is used for more than writing code.

docs/glossary.md
```diff
@@ -0,0 +1,265 @@
...
+
+#### Immature ticket
+
+Once tickets are mined they are immature for 256 blocks (about 20 hours) and cannot be called to vote until after this period elapses.
```
davecgh: I'm not sure if the specific values should be included in the glossary.  Due to the ability to change the consensus rules, the specific value might change in the future.  I think the point of the glossary is define the meaning of the terms as opposed to providing their exact values at a given point in time.

I'll refrain from commenting regarding the exact value point on the remaining entries, but I see this is recurring theme.

docs/glossary.md
```diff
@@ -0,0 +1,265 @@
...
+
+Transactions waiting to be mined.
+
+#### Miners
+
+Miners help secure the network via Proof-of-Work (PoW) mining. Miners recieve block rewards and transaction fees for successfully validating blocks. 
+
+#### Mining Pool
+
+A group of miners who share (pool) their computational resources to mine DCR. When a member of a mining pool successfully mines a block, the rewards are split among all pool members in proportion to the hashpower they contribute to the pool.
```
davecgh: I would suggest "the rewards are typically ...".  I say this, because the actual mechanism used varies by pool.

docs/glossary.md
davecgh: This definition is incorrect.  A block that is not built upon is not an orphan.  In the specific context this is referring to, it is a block that is part of a branch that has been abandoned.

Also, to expand on what @jrick said, "orphan" can also refer to a block that builds on top of another block that is missing as well as transactions that were in an orphaned block and have not been included in another one.

So, to summarize, orphans can refer to:

- Blocks that are a member of a branch of the blockchain that has been abandoned
- Blocks that build on an unknown block (the "parent" block is unknown making it an "orphan")
- Transactions with missing inputs (the "parent" transactions are unknown making them "orphans"
- Transactions that were part of a block that has been abandoned and have not been included in another block.  For example, a common case of this is for coinbase transactions and votes which are not allowed to reappear in another block.

docs/glossary.md
davecgh: Also, the DCR is not "stored" in an address.  Perhaps `An astronomically large secret number that allows DCR to be spent.  All transactions that spend coins must include a signature that can only produced with knowledge of the secret number."

docs/glossary.md
```diff
@@ -0,0 +1,265 @@
...
+
+See [consensus rules](#consensus-rules). 
+
+#### Public key
+
+A long string of alphanumeric characters generated algorithmically from a private key. The user’s public key is used to prove that a transaction was signed using their private key.
+
+#### Quorum
+
+The minimum level of participation required in order for a decision-making process to produce a valid outcome. Changes to the consensus rules require at least 10% of votes to be for or against the change in order to be valid. Politeaia proposals require 20%.
```
davecgh: Perhaps replace "valid" with "actionable"?  The outcome is not really invalid if it doesn't reach quorum, rather, it's just ignored.  Once again, I would avoid specific values that can change over time.

docs/glossary.md
davecgh: No, the chain with the most cumulative work wins.  Neither the number of tickets (so long as there are at least the minimum number required to satisfy the consensus rules that is), nor the number of blocks determine the ultimate winning branch (chain).

It is true that the number of tickets influence miner behavior, and the result is that there is strong preference towards creating blocks with more tickets, but that is something different.

docs/glossary.md
s-ben: Have split this into two definitions. Got the orphaned transaction definition from 'Mastering Bitcoin', Andreas Antonopoulos' book, which is available in its entirety on Google Books it appears. I think he wouldn't mind:) 

#### Orphan block

Orphaned blocks are valid blocks which are not included in the definitive blockchain. They can occur naturally when two miners produce blocks at similar times. They can also be caused by an attacker attempting to reverse transactions in a 51% attack. 


#### Orphan transaction

A transaction with inputs referring to a transaction that is not yet known, such as a missing parent.

docs/glossary.md
s-ben: Changed to below. 

#### Hash function

A cryptographic function that produces a fixed-size and unique hash value from variable-size input. Decred's Proof-of-Work uses the[BLAKE-256](https://docs.decred.org/research/blake-256-hash-function/) hashing function.

docs/glossary.md
s-ben: @davecgh , have updated definitions to include the extra info you provided (below). This is more of a developer term than an average user term, and I think developers would want a more detailed definition.

#### Orphan block

Orphaned blocks are valid blocks which are not included in the definitive blockchain. Orphan blocks can occur when they are part of a branch of the blockchain that has been abandoned. This can occur naturally when two miners produce blocks at similar times. Orphan blocks can also be created when they build on an unknown block (i.e. the "parent" block is unknown, making it an "orphan"). 

#### Orphan transaction

A transaction with missing inputs (i.e. the "parent" transaction is unknown, making the transaction an "orphan"). Opphan transactions can also be created when they are part of a block that has been abandoned and have not been included in another block. For example, if a miner creates a valid block with transactions specific to them, such as coinbase transactions and votes, and that block is orphaned, the transactions in that block will become orphaned transactions.

docs/glossary.md
s-ben: Going with @davecgh 's definition:

#### Private key

An astronomically large secret number that allows DCR to be spent. All transactions that spend DRC must include a signature that can only be produced with knowledge of this secret number.

docs/glossary.md
s-ben: Reworded:

#### Private passphrase

A passphrate that is used to encrypt parts of the wallet.db file; most notably private keys. A wallet.db file encrypted with a private passphrase cannot be used without that passphrase.

docs/glossary.md
s-ben: Is "cryptographics code" (term taken from another glossary's definition) an appropriate replacement for "string of characters"? If not, what do we call it? Rewording below. 

#### Public key

A cryptographic code generated algorithmically from a private key. The user’s public key is used to prove that a transaction was signed using their private key.

docs/glossary.md
s-ben: Reworded below. I am wondering though was "work" means, precisely. My googling has only turned up threads with developers using the term without explaining it. Is it just the sum of all difficulties in each block for the whole chain? Do we need to define work here (or elsewhere)?

#### Reorg

A reorganization (or reorg) of the blockchain is when a set of blocks are replaced by another set which has more work. The number of blocks replaced is the depth of the reorg.

docs/glossary.md
s-ben: Rewording:

#### Seed

A 256-bit (32-byte) cryptographically secure random number that can be used to recover a wallet. When creating a wallet using any Decred wallet software, a 33 word "seed phrase" is also created from the 32-byte seed. If the wallet is lost or corrupted, the seed phrase can be used to restore the wallet's private keys, transaction history, and ballances using any Decred wallet on any computer.

docs/glossary.md
s-ben: Added 'typically' to definition. 

#### Mining Pool

A group of miners who share (pool) their computational resources to mine DCR. When a member of a mining pool successfully mines a block, the rewards are typically split among all pool members in proportion to the hashpower they contribute to the pool.

docs/glossary.md
s-ben: Replaced valid with actionable.

#### Quorum

The minimum level of participation required in order for a decision-making process to produce an actionable outcome. Changes to the consensus rules require at least 10% of votes to be for or against the change in order to be valid. Politeaia proposals require 20%.

docs/glossary.md
s-ben: Agreed. Reworded with suggestion.

#### Hybrid PoW/PoS

In Decred, a [hybrid PoW/PoS system](https://docs.decred.org/research/hybrid-design/) is used, whereby blocks mined by PoW miners must be approved by ticket holders, which are pseudorandomly selected from the ticket pool. This provides a check on PoW miners and increases the overall cost of attacking the network. When a block is approved, 60% of the block reward goes to the PoW miner, 30% goes to the holders of  tickets called to vote (approve blocks), and 10% goes to the Decred Treasury to fund the project.

docs/glossary.md
```diff
@@ -0,0 +1,265 @@
...
+
+Tickets that reached the end of their window without being called to vote - these can be revoked, but do not grant a reward.
+
+#### Faucet
+
+A mechanism for obtaining free (testnet) tokens.
+
+#### Hash
+
+The output of a cryptographic hashing function that produces a fixed-size and unique hash value from variable-size input.
```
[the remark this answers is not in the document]
s-ben: Implemented suggestion.

#### Hash

The output of a cryptographic hashing function that produces a fixed-size and unique value from variable-size input.

docs/glossary.md
s-ben: Removed line about reversed and double spent.

#### Confirmation

A transaction is confirmed when it has been included in a block that has been verified by the network (added to the blockchain). Each additional block added to the blockchain reconfirms all transactions in all previous blocks. The number of times a transaction has been confirmed (i.e. the number of blocks that include the transaction) is used as a measure of confidence that the transaction will remain in the blockchain. For example. a wallet or other service may requre a certain number of confirmations before it considers a transaction to be valid.

s-ben:
@jrick , @davecgh have incorporated all of your comments. I think we're almost there, but my most recent edits (pasted in my response to your comments for convenience), should be given a final read through. 

@justinsantoro , have included a table of contents. It could maybe be made better with some formatting, but not sure we should keep it. The plain list of terms seems cleaner to me. Curious to hear other opinions.

docs/glossary.md
davecgh: That is better I think and much closer, but it's not really a hash of the public key either.  Perhaps: "The address is a human-readable representation of the public key receiving payment."

docs/glossary.md
davecgh: The rewording looks good.  To answer your question, work refers to the number of hashes that were required to produce the chain.  Since, the required difficulty changes periodically, the same number of blocks can have a different cumulative amount of total work due to differing requirements.

docs/glossary.md
davecgh: "(i.e. the number of blocks that include the transaction)" doesn't read correctly to me, as I read that as claiming the transaction is contained in multiple blocks.  Perhaps "the number of blocks added since the block that includes the transaction".

docs/glossary.md
s-ben: Using that:

#### Address

A secure identifier that represents a possible destination for a payment, similar to an email address. The address is a human-readable representation of the public key receiving payment.

docs/glossary.md
davecgh: Look good.  I see a typo.  s/Opphan/Orphan/

s-ben: Fixed

docs/glossary.md
s-ben: Had thought of that language, but as jrick pointed out above, the most recent block is said to have 1 confirmation even though there are no blocks following it, thus introducing quite the word puzzle. I gave up trying to solve this puzzle and just tried to express this as an example. 

#### Confirmation

A transaction is confirmed when it has been included in a block that has been verified by the network (added to the blockchain). Each additional block added to the blockchain reconfirms all transactions in all previous blocks. The number of times a transaction has been confirmed is used as a measure of confidence that the transaction will remain in the blockchain. A wallet or other service may require a certain number of confirmations before it considers a transaction to be valid. For example, if a wallet requires three confirmations to consider a transaction valid, it would require the next two blocks after the block including the transaction to confirm the transaction.

docs/glossary.md
```diff
@@ -0,0 +1,344 @@
+## A
```
jholdstock: Could you add a title to the page please, similar to line 1 in the other .md files in the repo.

I know mkdocs adds a top level header automatically, but specifying our own maintains consistency with all the other pages, and also allows us to add an icon to the header (usually supplied by @mariapleshkova or @linnutee)

mkdocs.yml
```diff
@@ -110,4 +110,5 @@ nav:
 - About:
   - 'Credits': 'about/credits.md'
   - 'License': 'about/license.md'
+- 'Glossary': 'glossary.md'
```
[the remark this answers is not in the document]
jholdstock: Nope - I am wrong about this. Not necessary.

docs/glossary.md
s-ben: Added title. Just going with no icon for now, similar to the Home page. But would be cool to have an icon.

Also, put a line below the title, as also appears standard on the other pages.

docs/glossary.md
```diff
@@ -0,0 +1,360 @@
+# Glossary
+
+---
+
+## A
```
jholdstock:
```suggestion
## A
---
```

docs/glossary.md
jholdstock: Typo here: passphrate => passphrase

docs/glossary.md
jholdstock: I think here we can go with:

> **Public key**
> An astronomically large number generated algorithmically from a private key. Public keys can be freely shared with the world without revealing any information about the private key they are generated from. The user’s public key is used to prove that a transaction was signed using their private key.

Thoughts @jrick @davecgh ?

jholdstock:
I've taken another look and resolved almost all of the issues above. I will aggregate the outstanding ones here:

- Add a linebreak under the A heading (I added a suggestion above, click to merge)
- Typos:
    - passphrate
    - ballance
- I've suggest a definition for public key above. Requesting jrick/davecgh to review
- davecgh has made a point about including specific values (quorum numbers, ticket maturity period)
    - Personally I feel like listing these numbers is acceptable. I definitely have no issue listing the ticket maturity period as this seems very unlikely to change. 

Based on comments above I have also opened new issues for:
- Add the term "Work" #702
- Add SPV vs fully validating #703
- Add the term "signature" #704

s-ben:
@jholdstock , have implemented your requested changes (fixed typos, added linebreak, added 'public key' definition). I went ahead and speculatively implemented 'public key' definition because I think it's good and unlikely to contain any inaccurate information (though still waiting on final review).  I did remove the "with the world" part, for conciseness (can add back if you like). Definition (awaiting review) is now: 

> #### Public key

> An astronomically large number generated algorithmically from a private key. Public keys can be freely shared without revealing any information about the private key they are generated from. The user’s public key is used to prove that a transaction was signed using their private key.

Assuming the separate issues you're opened can be addressed after we publish, it looks like all that's left is review of the 'public key' definition by @davecgh / @jrick , and decision on whether to include specific values (quorum numbers, ticket maturity period). It seems there is rough consensus to include these values. However, one compromise could be to remove any numbers that could reasonably change in the near to medium term. Working as a technical writer, I've definitely seen how easily numbers can go out of date. E.g. perhaps the Pi quorum and approval numbers could change in the near future as we tweak Pi? Or ticket maturity period?

docs/glossary.md
s-ben: Fixed typo

docs/glossary.md
s-ben: Fixed

docs/glossary.md
```diff
@@ -0,0 +1,361 @@
...
+---
+
+#### Account
+
+Accounts allow you to keep separate records of your DCR funds. Transferring DCR across accounts will create a transaction on the blockchain.
+
+#### Address
+
+A secure identifier that represents a possible destination for a payment, similar to an email address. The address is a human-readable representation of the public key receiving payment.
```
jrick: Addresses are human-readable forms of standard output script types, the most common one being Pay-to-Pubkey-Hash (P2PKH).  In this address format, the address is created by the hash of the public key, not the public key itself, and the redeemer must be able to provide a public key that hashes to a matching hash.  Other types of address formats include Pay-to-Pubkey (P2PK) which works as described here currently, but their usage is not very common, as well as Pay-to-Script-Hash (P2SH) in which more complex and non-standard redeeming requirements can be provided by the redeemer as long as they provide a nested script which hashes to the expected value.

docs/glossary.md
```diff
@@ -0,0 +1,361 @@
...
+
+A tool for inspecting the contents of blocks in a more user-friendly format. 
+
+#### Block header
+
+Metadata at the beginning of a block that defines information about the block. The hash of a block header is the main way of identifying a block. 
+
+#### Block reward
+
+A reward (in DCR) split between PoW miners, ticket holders, and the Treasury for successfully approving a block of transactions. The block reward [decays exponentially](https://docs.decred.org/advanced/inflation/) with block height (length of blockchain). However, Decred’s algorithm interpolates this decay over time so as not to produce market shocks, as typically seen in algorithms that involve large and infrequent reductions, such as halving the rewards every 4 years. The block reward started at 31.19582664 and it adjusts every 6,144 blocks (approximately 21.33 days) by reducing by a factor of 100/1011.
```
jrick: > A reward (in DCR) split between PoW miners, ticket holders,

Instead of saying ticket holders here, maybe say voters.  Only voters who are picked for this block and whose votes appear in the block are rewarded.

> for successfully approving a block of transactions.

Both PoW miners and stake voters are rewarded even if the stake voters collectively vote to disapprove of regular transactions in the prior block.

docs/glossary.md
```diff
@@ -0,0 +1,361 @@
...
+
+#### Block voting
+
+In each block, five tickets are called to vote. In addition to votes on any open consensus rule change proposals, each ticket votes to approve or reject the regular transaction tree of the previous block. If a majority of voting tickets vote No, the regular transaction tree of the previous block is rejected and those transactions are returned to the mempool.
```
jrick: > If a majority of voting tickets vote No,

A majority must vote yes to approve, otherwise this results in disapproval (including if there is a tie).

docs/glossary.md
```diff
@@ -0,0 +1,361 @@
...
+Tickets that reached the end of their window without being called to vote - these can be revoked, but do not grant a reward.
+
+
+## F
+
+---
+
+#### Faucet
+
+A mechanism for obtaining free (testnet) tokens.
```
jrick: Would like to see "coins" used here rather than "tokens".  Tokens generally refer to a second layer currency on top of another network.

docs/glossary.md
```diff
@@ -0,0 +1,361 @@
...
+
+
+## H
+
+---
+
+#### Hash
+
+The output of a cryptographic hashing function that produces a fixed-size and unique value from variable-size input.
```
jrick: Hashes are not unique, but good hash functions are astronomically difficult to reverse or to find collisions.

https://en.wikipedia.org/wiki/Pigeonhole_principle

docs/glossary.md
```diff
@@ -0,0 +1,361 @@
...
+
+A cryptographic function that produces a fixed-size and unique hash value from variable-size transaction input. Decred's Proof-of-Work uses the [BLAKE-256](https://docs.decred.org/research/blake-256-hash-function/) hashing function. 
+
+#### Hashrate
+
+The number of hashes per second computed by miners on the network.
+
+#### Hybrid PoW/PoS
+
+In Decred, a [hybrid PoW/PoS system](https://docs.decred.org/research/hybrid-design/) is used, whereby blocks mined by PoW miners must be approved by ticket holders, which are pseudorandomly selected from the ticket pool. This provides a check on PoW miners and increases the overall cost of attacking the network. When a block is approved, 60% of the block reward goes to the PoW miner, 30% goes to the holders of  tickets called to vote (approve blocks), and 10% goes to the Decred Treasury to fund the project.
```
jrick: This section should define the terms Pow and PoS as Proof-of-Work and Proof-of-Stake respectively.

docs/glossary.md
```diff
@@ -0,0 +1,361 @@
...
+#### Immature ticket
+
+Once tickets are mined they are immature for 256 blocks (about 20 hours) and cannot be called to vote until after this period elapses.
+
+#### Inflation
+
+Increase in the [available supply](https://docs.decred.org/advanced/inflation/) of Decred as new DCR is minted into existence through the block reward.
+
+
+## I
```
jrick: Wrong letter, you want an L here.

docs/glossary.md
```diff
@@ -0,0 +1,361 @@
...
+
+Miners help secure the network via Proof-of-Work (PoW) mining. Miners recieve block rewards and transaction fees for successfully validating blocks. 
+
+#### Mining Pool
+
+A group of miners who share (pool) their computational resources to mine DCR. When a member of a mining pool successfully mines a block, the rewards are typically split among all pool members in proportion to the hashpower they contribute to the pool. 
+
+#### Missed ticket
+
+Tickets that have been called but did not receive a reward. This can happen if a ticket is called to vote and does not respond. This can also happen if a ticket is called to vote, responds and broadcasts their vote to the network, but a miner does not include the vote in the following block.
```
jrick: This would read better if it referred to voting wallets watching for called tickets and voting when possible, rather than tickets themselves responding to votes.

docs/glossary.md
```diff
@@ -0,0 +1,361 @@
...
+
+---
+
+#### Politeia
+
+A system for facilitating the submission and discussion of [proposals](https://proposals.decred.org/) in an environment with transparent censorship. 
+
+#### Private key
+
+An astronomically large secret number that allows DCR to be spent. All transactions that spend DRC must include a signature that can only be produced with knowledge of this secret number.
```
jrick: DRC -> DCR

> All transactions ... must include a signature that can only be produced with knowledge of this secret number.

This is not universally true.  The only requirement is that the script virtual machine is able to execute the scripts in the input and output without failure.  There is no requirement that a signature check is performed, but this is the most common way to protect DCR from being spent by unauthorized parties.  One example of spending requirements that does not perform any signature validation is a puzzle, in which case any arbitrary data can be provided and so long as it hashes to the expected value, the script executes and the output can be spent.

docs/glossary.md
```diff
@@ -0,0 +1,361 @@
...
+
+#### Private passphrase
+
+A passphrase that is used to encrypt parts of the wallet.db file; most notably private keys. A wallet.db file encrypted with a private passphrase cannot be used without that passphrase.
```
jrick: The private keys can not be accessed unless the passphrase is used to decrypt them.  Other unencrypted data saved to the DB is always accessible.

docs/glossary.md
s-ben: This is really good info for a general definition of addresses. But it's not clear to me from this comment which specific method Decred addresses use. Do we use P2PKH? P2PK? iirc, this definition was iterated on in a previous review, and we came up with this, which is accurate but high-level and concise. I think it would be valuable to say technically how Decred generates its hash. A full paragraph about addresses in general is useful as well to the general crypto community, if we don't mind this definition being a bit long. When researching definitions for this glossary, I found that the definitions found in the top glossaries on google were quite lacking, in detail and accuracy.

davecgh: All of them are supported and used.  The most common is P2PKH (starts with Ds), but most VSPs use P2SH (starts with Dc) and the configuration of them uses P2PK (starts with Dk).

Furthermore, there is support for (but nothing really uses it yet) Schnorr signature variants that are different still.

That said, for a glossary, I'm not sure it should go into the specific technical implementation details that are subject to change over time.  Rather, I personally think it should stick to something that more accurately describes their intended purpose as opposed to how they specifically carry it out.  The existing text, I think is pretty close with the exception of the not exactly accurate bits about the public key.

Perhaps:

A human-readable representation of a possible destination for a payment, similar to an email address, with the exception that addresses are typically only intended for a single use.  There are variety of address types that may be used to represent a wide range of constraints required to redeem funds such as requiring a signature from a single party, signatures from multiple parties, or a specific minimum time in the future.

docs/glossary.md
s-ben: Have taken a stab at a new definition:

> A reward (in DCR) split between PoW miners, voters (ticket holders that have been called to vote), and the Treasury for successfully creating a block of transactions. 

For the second comment, I wasn't sure what language to use. I've just replaced 'approving' with 'creating', in hopes that works. After reading again the [hybrid-design](https://docs.decred.org/research/hybrid-design/) page, I have a better understanding. But the temporal logic still eludes me... Can you think of a better way to word this?

docs/glossary.md
s-ben: Updated to. Below. My only question (as a reader) is, if there are five votes, how is a tie possible? Can a ticket abstain from voting to approve transactions? In the docs, I see tickets can abstain vote on consensus rules, but I'm not seeing any reference to abstaining from voting on transactions. 

#### Block voting

In each block, five tickets are called to vote. In addition to votes on any open consensus rule change proposals, each ticket votes to approve or reject the regular transaction tree of the previous block. If a majority of voting tickets vote Yes, the regular transaction tree of the previous block is accepted. If a majority of voting tickets vote No (or there is a tie), the regular transaction tree of the previous block is rejected and those transactions are returned to the mempool.

docs/glossary.md
s-ben: Implemented suggestion

docs/glossary.md
jrick: There can be four votes.  Only three votes are necessary to mine a block.

docs/glossary.md
s-ben: Interesting. Good to know. Updated definition below. I've also updated the 'Hash Function' definition accordingly (also below).

#### Hash

The output of a cryptographic hashing function that produces a fixed-size value from variable-size input. While it is computationally easy to create a hash from an input, it is extremely computationally difficult to calculate an input that will produce a given hash. 


#### Hash function

A cryptographic function that produces a fixed-size hash value from variable-size transaction input. Decred's Proof-of-Work uses the [BLAKE-256](https://docs.decred.org/research/blake-256-hash-function/) hashing function.

docs/glossary.md
s-ben: Agreed. I think though that we should punt on those for now, and include those definitions in the next version. @jholdstock has already broken out the definition for "work" (which is PoW?) into a separate Issue on dcrdocs,
https://github.com/decred/dcrdocs/issues/702

We're doing the same with "Signature",
https://github.com/decred/dcrdocs/issues/704

My concern is that we could continue iterating on this forever, which keeps the accurate and useful definitions from the community in the meantime.

docs/glossary.md
s-ben: Nice catch. Fixed

docs/glossary.md
s-ben: Agreed. Reworded to:

#### Missed ticket

Tickets that have been called but did not receive a reward. This can happen if a ticket is called to vote, but the wallet that bought the ticket does not respond. This can also happen if the wallet does respond and broadcasts its vote to the network, but a miner does not include their vote in the following block.

docs/glossary.md
s-ben: Googled around trying to understand this better. Couldn't come up with a wording that keeps the level of detail we have here and is still accurate. Have retreated to a higher level of abstraction (below). If you can think of more useful wording, can use that.

#### Private key

An astronomically large secret number that allows DCR to be spent. All transactions that spend DCR can only be produced with knowledge of this secret number.

docs/glossary.md
s-ben: Reworded to:

#### Private passphrase

A passphrase that is used to encrypt parts of the wallet.db file; most notably private keys. The private keys can not be accessed unless the passphrase is used to decrypt them.

docs/glossary.md
jrick: Would go with:

> An astronomically large number which, when kept secret from others, enables legitimate spending of personal Decred while preventing others from spending it.  Private keys are encrypted by wallets to prevent theft.

s-ben: Implemented suggestion.

jholdstock:
Sorry to drag this out further but I've noticed one more thing.

Internal links in this new document should be relative paths to the relevant `.md` file. At the moment they are links to `https://docs.decred.org/whatever`

I'm counting 9 links which need to be updated

s-ben:
Nah, good catch. Absolute link will just create more work in the long run. Have updated to relative links and pushed.

docs/glossary.md
s-ben: After some schooling by @jrick in #documentation yesterday, I feel I mostly understand he above. However, so far the glossary definitions are fairly high-level and user-centric, which is IMO appropriate for dcrdocs. Going too much into the weeds here will confuse users that just want to know how an address works in Decredition, etc. I'm happy to debate, but I would vote for publishing this definition as-is for now and moving debate to this Issue @jholdstock has created.
https://github.com/decred/dcrdocs/issues/710

I do think this is valuable information for a developer. Have copied the above comments to my notes for use in more developer-facing docs.

docs/glossary.md
s-ben: Have gone with your suggested rewording. Fixed the reduction factor. Holding off for now on adding mention about possibly not getting block reward. This seems a little detailed for the glossary, and @RichardRed0x I believe is adding a mention or two about this in the big Jargon update PR: https://github.com/decred/dcrdocs/pull/590

s-ben:
Have addressed two remaining comments. If @jrick and @davecgh are cool with my suggestions, I think we're good to merge and iterate in separate Issues.

PR #695, opened 2018-11-12
Removes all references to web wallet

### LINKED ISSUE #694: Remove references to web wallet
Per discussion on Issue #688 , there is an opinion that references to the web wallet should be removed. Does anyone want to keep the web wallet?

Comments:

@dajohi 

> all references to web wallet should be removed.

@jholdstock 
> The web wallet was put online for airdrop recipients when the only other available wallet was dcrwallet via CLI. It is not a very user friendly wallet, it is made practically redundant by Decrediton (especially with SPV mode). It is a point of centralisation which requires the user to trust a remote server which is not under their own control. The code is a nightmare. There is time/cost associated with running the web wallet.

s-ben:
The web wallet has been deprecated and will be taken offline soon. It is also insecure, and we don't want new users of it. This PR removes all mentions of the web wallet ( Closes #694 ).


s-ben:
Removed `Redirect /getting-started/user-guides/web              /wallets/web` from `httpd.conf`.

PR #709, opened 2018-11-22
Update to decred_supply image

zubairzia0:
Changed the following:
- Dec Subsidy to Decred Treasury
- Adjust font size to 9.5
-Font colour is now black
- Added tickmarks on the x and y axis.


jholdstock:
Looks good to me but the PR needs to be rebased

zubairzia0:
does this work?

PR #726, opened 2018-12-03
Add guide for SPV.

### LINKED ISSUE #676: SPV
SPV deserves its own page. Material we could cover:
- why it was added to decred
- how it works.
- the pro/cons of using it.
- privacy/security implications. 
- how SPV in decred differs from other SPV implementations.

jholdstock:
Closes #676

### DIFF
--- a/docs/wallets/spv.md
+++ b/docs/wallets/spv.md
@@ -1,0 +1,60 @@
+# SPV (Simple Payment Verification)
+
+---
+
+## What is Simple Payment Verification?
+
+Simple Payment Verification (SPV) allows the use of a Decred wallet without having to download the entire Decred blockchain. A wallet operating in SPV mode is able to retrieve all of its transactions whilst only downloading blocks from the Decred network if they contained owned transactions. In a typical case a wallet will only need to download tens of megabytes rather than multiple gigabytes.
+
+SPV has been built directly into the `dcrwallet` CLI tool as a secondary operating mode. Decrediton uses `dcrwallet` behind the scenes, and as a result all users of Decred are able to enable SPV regardless of their wallet preference.
+
+
+## Why was SPV added to `dcrwallet`?
+
+The hardware requirements to run a Decred wallet are drastically reduced when operating in SPV mode. Storage and download requirements are reduced because rather than downloading the whole blockchain, an SPV wallet will only download the blocks which contain relevant transactions, and for all of the other blocks in the chain only the headers are downloaded. The amount of processing power required is reduced because a device running an SPV wallet will only validate the proof-of-work and the header chain rather than validating every transaction in every block and ensuring the block contents are consistent with the headers.
+
+As a result of these decreased requirements, Decred wallets can operate on a wider set of devices, particularly mobile devices. Smartphones and tablets are typically limited in at least one of CPU power, storage or download capacity, which makes running a full node either impractical or impossible. 
+
+Another benefit offered by SPV is an extreme reduction in the time required for a brand new wallet to become operational, offering a huge improvement to the user experience.
+
+
+## How does SPV work?
+
+At the start of every block added to the Decred blockchain is 180 bytes of data called the [block header](../advanced/block-header-specifications.md). The block header describes key information about the block including the hash of the block, the merkle root (the sum of all the transaction hashes in the block), and the nonce calculated by the proof-of-work miners. A predetermined filter is also created for every block, based on the all of the transactions within the block. 
+
+When an SPV wallet initialises it will connect to the Decred network using peer-to-peer connections, and it will download the full set of headers and filters. It will then validate the header chain to ensure that the chain and its proof-of-work are valid. Once this is complete, the wallet will use the filters to locally identify which blocks contains owned transactions without uploading any private data to remote nodes. The wallet can then use the peer-to-peer network to download these blocks, scan them for relevant transactions and select these to update personal transaction history and balance. 
+
+
+## How is this different from a "light" wallet?
+
+"Light" wallets such as [Exodus](https://www.exodus.io/) or [Atomic](https://atomicwallet.io/) already bring some of the benefits of an SPV wallet - for example these wallets allow sending and receiving Decred transactions with a minimal start-up time and without downloading the full chain. Light wallets certainly provide a level of convenience but there are often subtle consequences because they depend upon an API provided by a centralised service:
+
+- Light wallets depend on a remote service to watch owned addresses and provide notifications when transactions are received by the addresses. Notifications could be missed or incorrect.
+
+- Users will often be required to upload an extended public key to the central server so it is aware of all of owned addresses, potentially compromising the users privacy.
+
+- Light wallets do not validate the information they receive by checking the blockchain directly - they have to blindly trust the information provided by the central server.
+
+These concerns do not apply to SPV wallets because they connect directly to the decentralised peer-to-peer Decred network, they upload no private data to remote nodes, and they validate their own transactions by inspecting the blockchain directly.
+
+
+## Are there any disadvantages?
+
+- SPV does not support voting wallets. Voting wallets have the responsibility to vote on the validity of the last block, and a wallet cannot be sure of the validity unless it fully validates the whole blockchain leading up to that block. It is possible however to purchase tickets and allocate the voting rights to a [Voting Service Provider](../mining/how-to-stake.md#pos-using-a-voting-service-provider-vsp).
+
+- SPV wallets only download blocks which have transactions related to their owned addresses, which could potentially reveal more information about the wallet than if it downloaded every single block. This only presents a very minor decrease in privacy, but it is a decrease nonetheless. This is mitigated by downloading blocks from multiple peers so no single peer is able to see the full list of blocks downloaded by a wallet. Even if a passive observer on the network is able to see which blocks are downloaded by a wallet, they are not able to identify which transactions in those blocks are relevant.
+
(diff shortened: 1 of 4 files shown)


docs/wallets/spv.md
```diff
@@ -0,0 +1,60 @@
+# SPV (Simple Payment Verification)
+
+---
+
+## What is Simple Payment Verification?
+
+Simple Payment Verification (SPV) allows the use of a Decred wallet without having to download the entire Decred blockchain. A wallet operating in SPV mode is able to retrieve all of its transactions whilst only downloading blocks from the Decred network if they contained owned transactions. In a typical case a wallet will only need to download tens of megabytes rather than multiple gigabytes.
```
jrick: contained -> contain

docs/wallets/spv.md
```diff
@@ -0,0 +1,60 @@
...
+
+SPV has been built directly into the `dcrwallet` CLI tool as a secondary operating mode. Decrediton uses `dcrwallet` behind the scenes, and as a result all users of Decred are able to enable SPV regardless of their wallet preference.
```
jrick: maybe "all users of official wallets" instead of "all users of Decred".

docs/wallets/spv.md
```diff
@@ -0,0 +1,60 @@
...
+
+
+## Why was SPV added to `dcrwallet`?
+
+The hardware requirements to run a Decred wallet are drastically reduced when operating in SPV mode. Storage and download requirements are reduced because rather than downloading the whole blockchain, an SPV wallet will only download the blocks which contain relevant transactions, and for all of the other blocks in the chain only the headers are downloaded. The amount of processing power required is reduced because a device running an SPV wallet will only validate the proof-of-work and the header chain rather than validating every transaction in every block and ensuring the block contents are consistent with the headers.
+
+As a result of these decreased requirements, Decred wallets can operate on a wider set of devices, particularly mobile devices. Smartphones and tablets are typically limited in at least one of CPU power, storage or download capacity, which makes running a full node either impractical or impossible. 
+
+Another benefit offered by SPV is an extreme reduction in the time required for a brand new wallet to become operational, offering a huge improvement to the user experience.
```
jrick: mobile operating systems also greatly reduce the amount of background processing apps can perform, making running a full node unreliable and slow to use again after it was not running for a while.  The same problem can affect SPV wallets, but the sync to catch up is far faster.  This ties in with the previous paragraph about mobile app restrictions.  Not sure if you think this is important to detail here, but it is what it is.

docs/wallets/spv.md
```diff
@@ -0,0 +1,60 @@
...
+
+- Light wallets depend on a remote service to watch owned addresses and provide notifications when transactions are received by the addresses. Notifications could be missed or incorrect.
+
+- Users will often be required to upload an extended public key to the central server so it is aware of all of owned addresses, potentially compromising the users privacy.
+
+- Light wallets do not validate the information they receive by checking the blockchain directly - they have to blindly trust the information provided by the central server.
+
+These concerns do not apply to SPV wallets because they connect directly to the decentralised peer-to-peer Decred network, they upload no private data to remote nodes, and they validate their own transactions by inspecting the blockchain directly.
```
jrick: The only validation performed is the proof-of-work check. Transactions contained in the block may still be invalid, but it is unlikely (and economically unreasonable) to spend PoW resources on mining an invalid chain.

docs/wallets/spv.md
```diff
@@ -0,0 +1,60 @@
...
+
+
+## Are there any disadvantages?
+
+- SPV does not support voting wallets. Voting wallets have the responsibility to vote on the validity of the last block, and a wallet cannot be sure of the validity unless it fully validates the whole blockchain leading up to that block. It is possible however to purchase tickets and allocate the voting rights to a [Voting Service Provider](../mining/how-to-stake.md#pos-using-a-voting-service-provider-vsp).
+
+- SPV wallets only download blocks which have transactions related to their owned addresses, which could potentially reveal more information about the wallet than if it downloaded every single block. This only presents a very minor decrease in privacy, but it is a decrease nonetheless. This is mitigated by downloading blocks from multiple peers so no single peer is able to see the full list of blocks downloaded by a wallet. Even if a passive observer on the network is able to see which blocks are downloaded by a wallet, they are not able to identify which transactions in those blocks are relevant.
```
jrick: The current SPV implementation is multi-peer, but it requests relevant blocks from every connected peer rather than splitting these requests up among various distinct peers.  This can be changed, but users should not be mislead into believing this mitigation is currently being performed.

docs/wallets/spv.md
```diff
@@ -0,0 +1,60 @@
...
+
+- Wallets operating in SPV mode are only able to validate the block headers they download and not the filters. This makes a "false-negative" attack possible: a malicious peer which knows a wallet is waiting for a particular transaction could send the wallet a fake filter which does not include the transaction, resulting in the wallet not downloading the block and so not becoming aware of the transactions existence. This transaction would still be visible to all fully validating nodes and wallets, and it will still appear in the [block explorer](../getting-started/using-the-block-explorer.md). One way to prevent this vulnerability is to add the hash of the filter into the block header, enabling SPV wallets to easily check the validity of the filters without having to download their blocks. A "false-positive" scenario is not possible - if a malicious node provides a fake filter which includes a non-existent transaction, the wallet will simply download the full block, compare it to the filter and discover that the filter is not genuine.
```
jrick: It would be worth noting here that there are preliminary proposals to solving this issue by committing the filter to the part of the block header that is PoW-validated.

docs/wallets/spv.md
```diff
@@ -0,0 +1,60 @@
...
+
+
+## How do I use SPV?
+
+#### `dcrwallet` CLI
+
+To enable SPV mode in `dcrwallet` simply provide the `--spv` flag when starting the process. There is also an optional `--spvconnect` flag which allows you to specify the IP of a full node you wish to sync from.
```
jrick: --spvconnect can also be specified multiple times to sync from multiple peers.  Additionally, DNS seeding is disabled with --spvconnect and the wallet will only connect to these peers rather than targeting a total of 8 outbound connections.

docs/wallets/spv.md
jholdstock: So just to clarify, if a wallet is connected to only one peer, it will download every block from that peer?

jrick: The ones that match the filter, but yes.

docs/wallets/spv.md
```diff
@@ -0,0 +1,60 @@
...
+
+
+## How does SPV work?
+
+At the start of every block added to the Decred blockchain is 180 bytes of data called the [block header](../advanced/block-header-specifications.md). The block header describes key information about the block including the hash of the block, the merkle root (the sum of all the transaction hashes in the block), and the nonce calculated by the proof-of-work miners. A predetermined filter is also created for every block, based on the all of the transactions within the block. 
+
+When an SPV wallet initialises it will connect to the Decred network using peer-to-peer connections, and it will download the full set of headers and filters. It will then validate the header chain to ensure that the chain and its proof-of-work are valid. Once this is complete, the wallet will use the filters to locally identify which blocks contains owned transactions without uploading any private data to remote nodes. The wallet can then use the peer-to-peer network to download these blocks, scan them for relevant transactions and select these to update personal transaction history and balance.
```
s-ben: contains->contain

docs/wallets/spv.md
```diff
@@ -0,0 +1,60 @@
...
+
+- Light wallets depend on a remote service to watch owned addresses and provide notifications when transactions are received by the addresses. Notifications could be missed or incorrect.
+
+- Users will often be required to upload an extended public key to the central server so it is aware of all of owned addresses, potentially compromising the users privacy.
+
+- Light wallets do not validate the information they receive by checking the blockchain directly - they have to blindly trust the information provided by the central server.
+
+These concerns do not apply to SPV wallets because they connect directly to the decentralised peer-to-peer Decred network, they upload no private data to remote nodes, and they discover their own transactions by inspecting the blockchain directly.
```
s-ben: Runon sentence. Suggest reword to:

> These concerns do not apply to SPV wallets because they connect directly to the decentralised, peer-to-peer Decred network. They upload no private data to remote nodes and they discover their own transactions by inspecting the blockchain directly.

docs/wallets/spv.md
```diff
@@ -0,0 +1,60 @@
...
... download and not the filters. This makes a "false-negative" attack possible: a malicious peer which knows a wallet is waiting for a particular transaction could send the wallet a fake filter which does not include the transaction, resulting in the wallet not downloading the block and so not becoming aware of the transactions existence. This transaction would still be visible to all fully validating nodes and wallets, and it will still appear in the [block explorer](../getting-started/using-the-block-explorer.md). One way to prevent this vulnerability is to add the hash of the filter into the part of the block header that is PoW validated, enabling SPV wallets to easily check the validity of the filters without having to download their blocks. A proposal to make this change has [already been suggested](https://github.com/decred/dcrd/issues/971), however a hardfork will be required to make the required change to the block header format.  A "false-positive" scenario is not possible - if a malicious node provides a fake filter which includes a non-existent transaction, the wallet will simply download the full block, compare it to the filter and discover that the filter is not genuine.
```
s-ben: Dubious use of ':'. Suggest rewording to:

> This makes a "false-negative" attack possible, whereby a malicious peer that knows a wallet is waiting for a particular transaction could send the wallet a fake filter which does not include the transaction, resulting in the wallet not downloading the block and so not becoming aware of the transactions existence.

s-ben: Dubious use of '-'. Suggest rewording to:

> A "false-positive" scenario is not possible. If a malicious node provides a fake filter which includes a non-existent transaction, the wallet will simply download the full block, compare it to the filter and discover that the filter is not genuine.

jholdstock:
Thanks @jrick and @s-ben, I've made the changes suggested above.

s-ben:
Looks good to me @jholdstock. I would say it's ready to merge.

jholdstock:
Rebased onto master. I think this is good to go if @jrick approves the changes on his comments.

PR #764, opened 2018-12-21
Updates transaction details page

### LINKED ISSUE #760: Improve transaction details page
Mentioned by dave on slack

> I noticed in support last night that https://docs.decred.org/advanced/transaction-details/ is slightly out of date
> 
> For one, the serialization formats at the bottom need to remove 3 and 4
> https://github.com/decred/dcrd/blob/master/wire/msgtx.go#L136-L151
> 
> Those should've never been included in there as they have to do with the signature hash, not wire serialization
> 
> Also, it doesn't call out how the version is treated as a version and the serialization type
> 
> I would also highly suggest showing an example encode/decode, like https://docs.decred.org/advanced/block-header-specifications/

s-ben:
This PR removes two unnecessary serialization formats and improves the definition of the `Version` parameter in the Transaction Format table ( Closes #760 ). 

Note: we also want to include an example transaction, but I ran out of time (headed traveling tomorrow). Have broken off the example transaction (including example tx provided by @davecgh) into a separate Issue: https://github.com/decred/dcrdocs/issues/765

### DIFF
--- a/docs/advanced/transaction-details.md
+++ b/docs/advanced/transaction-details.md
@@ -9,7 +9,7 @@ Decred transactions are transfers of DCR that exist within blocks. Transactions 
 
 Field        | Description                                                                                    | Size
 ---          | ---                                                                                            | ---
-Version      | Transaction version. This number is used to signify how the transaction should be interpreted. The first two bytes specify the version number. The second two bytes specify the serialization format.   | 4 bytes
+Version      | Transaction version. This number is used to signify how the transaction should be interpreted. The upper 16 bits specify the serialization format and the lower 16 bits specify the version number.   | 4 bytes
 Input count  | The number of inputs in the transaction encoded as a variable-length integer                   | 1-9 bytes
 Inputs       | Serialized list of all the transaction's inputs                                                | Variable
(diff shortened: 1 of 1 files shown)


docs/advanced/transaction-details.md
```diff
@@ -9,7 +9,7 @@ Decred transactions are transfers of DCR that exist within blocks. Transactions
 
 Field        | Description                                                                                    | Size
 ---          | ---                                                                                            | ---
-Version      | Transaction version. This number is used to signify how the transaction should be interpreted  | 4 bytes
+Version      | Transaction version. This number is used to signify how the transaction should be interpreted. The first two bytes specify the version number. The second two bytes specify the serialization format.   | 4 bytes
```
davecgh: I would say `The upper 16-bits specify the serialization format and the lower 16-bits specify the version number.`  

Whether or not it's the first or last 2 bytes depends on byte order while specifying them in terms of upper and lower bits does not.

s-ben: Changed to that wording. Though when I'm back I want to wrap my head around this bit shifting business so I can know my up from down :)

PR #782, opened 2019-01-12
Add new glossary items.

jholdstock:
- ASIC
- Coinbase maturity
- Coinbase transaction
- Cold wallet
- Full node
- Genesis block
- Hardware wallet
- Hot wallet
- Nonce
- Multisignature
- Stakebase transaction
- Tor

### DIFF
--- a/docs/glossary.md
+++ b/docs/glossary.md
@@ -13,6 +13,10 @@ Accounts allow you to keep separate records of your DCR funds. Transferring DCR 
 
 A human-readable representation of a possible destination for a payment, similar to an email address. Unlike an email address however, there are a variety of address types, and most addresses are intended only for a single use. This is because addresses represent not only the destination of a payment, but constraints required to redeem funds. For the most common type of address, the constraints require that the recipient provide both the public key which hashes to the requisite hash and a valid signature that is created from the corresponding private key. Other address types require signatures from multiple parties (multi-sig), signatures created within specified time limits, or other more complex constraints.
 
+#### ASIC
+
+An application specific integrated circuit (ASIC) is a computing device which has been designed to perform only a single task, but to perform that task with extreme efficiency. In the context of cryptocurrency, ASIC usually refers to a piece of hardware which is designed to perform proof-of-work mining. Hashrates from ASICs are typically orders of magnitudes higher than hashrates from general purpose computers.
+
 #### Atom
 
 The smallest unit of Decred currency. One atom is one hundred millionth of a single DCR (0.00000001 DCR).
@@ -46,6 +50,18 @@ In each block, five tickets are called to vote. In addition to votes on any open
 ## C
 
 ---
+
+#### Coinbase maturity
+
+Outputs from coinbase and stakebase transactions cannot be spent until the coinbase maturity period has passed. The coinbase maturity period is 256 blocks.
+
+#### Coinbase transaction
+
+Every block which is mined contains a single coinbase transaction. This transaction will only have one input, and that is newly created Decred which constitutes the PoW and Treasury portions of the block reward. Coinbase transactions are part of the regular transaction tree, which means that they will be rejected if proof-of-stake voters vote to reject the block which contains them. Decred created in coinbase transactions cannot be spend until the coinbase maturity period has passed.
+
+#### Cold wallet
+
+A Decred wallet which is used to generate private keys and and store Decred whilst being completely offline. Methods of cold storage include generating a wallet on a computer which is never connected to the Internet, using a dedicated hardware wallet, or using a paper wallet.
 
 #### Command Line Interface (CLI)
 
@@ -115,10 +131,27 @@ Tickets that reached the end of their window without being called to vote - thes
 
 A mechanism for obtaining free (testnet) coins.
 
+#### Full node
+
+An instance of the Decred daemon `dcrd` which is serving the full version of the blockchain to peers, and fully verifying all of the consensus rules of Decred. 
+
+
+## G
+
+---
+
+#### Genesis block
+
+The very first block of the Decred blockchain, created on the 8th of February 2016.
+
 
 ## H
 
 ---
+
+#### Hardware wallet
+
+A dedicated hardware device which securely stores a wallets private keys. Examples include [Trezor](https://trezor.io/) and [Ledger](https://www.ledger.com/).
 
 #### Hash
 
@@ -132,6 +165,10 @@ A cryptographic function that produces a fixed-size hash value from variable-siz
 #### Hashrate
 
 The number of hashes per second computed by miners on the network.
+
+#### Hot wallet
+
+A Decred wallet which is connected to the Internet. Proof of stake voting wallets are an example of hot wallets, they are constantly connected to the Decred network because they can be called to vote at any time.
 
 #### Hybrid PoW/PoS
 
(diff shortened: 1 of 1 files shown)


docs/glossary.md
```diff
@@ -13,6 +13,10 @@ Accounts allow you to keep separate records of your DCR funds. Transferring DCR
 
 A human-readable representation of a possible destination for a payment, similar to an email address. Unlike an email address however, there are a variety of address types, and most addresses are intended only for a single use. This is because addresses represent not only the destination of a payment, but constraints required to redeem funds. For the most common type of address, the constraints require that the recipient provide both the public key which hashes to the requisite hash and a valid signature that is created from the corresponding private key. Other address types require signatures from multiple parties (multi-sig), signatures created within specified time limits, or other more complex constraints.
 
+#### ASIC
```
s-ben: A little wordsmithing:

> An Application-Specific-Integrated-Circuit (ASIC) is a computing device which has been designed to perform only a single task with extreme efficiency. In the context of cryptocurrency, an ASIC is usually designed to perform Proof-of-Work (PoW) mining. Hashrates from ASICs are typically orders of magnitudes higher than hashrates from general purpose computers or GPUs.

s-ben:
@jholdstock these definitions look generally solid. Did some rewording around convention, grammar and flow.

jholdstock:
Thanks for that, I've included the improved definition of ASIC. I would appreciate an extra review for technical correctness before this is merged.

docs/glossary.md
```diff
@@ -323,6 +377,10 @@ The amount of DCR one must time-lock in order to buy a ticket. The ticket price
 
 The process of buying part of a ticket. This is done by coordinating with other parties who will buy the other parts of a ticket. This can be done without giving up custody to your DCR. The minimum amount of DCR that can be put into a split ticket is 5 DCR. Ticket-splitting is currently coordinated through the 'ticket-splitting' channel on the [Decred slack](https://slack.decred.org/) channel. 
 
+#### Tor
+
+The Onion Router (Tor) is free open-source software which enables users to improve security and privacy when communicating over a network. It is commonly used to protect against Internet traffic analysis or to enable anonymous communication.
+
```
dajohi: Would a link to https://www.torproject.org/ be useful here?

docs/glossary.md
```diff
@@ -47,6 +51,18 @@ In each block, five tickets are called to vote. In addition to votes on any open
 
 ---
 
+#### Coinbase maturity
+
+Outputs from coinbase and stakebase transactions cannot be spent until the coinbase maturity period has passed. The coinbase maturity period is 256 blocks.
+
+#### Coinbase transaction
+
+Every block which is mined contains a single coinbase transaction. This transaction will only have one input, and that is newly created Decred which constitutes the PoW and Treasury portions of the block reward. Coinbase transactions are part of the regular transaction tree, which means that they will be rejected if proof-of-stake voters vote to reject the block which contains them. Decred created in coinbase transactions cannot be spend until the coinbase maturity period has passed.
```
davecgh: This seems to be a recurring them, but I really don't think the Glossary should be calling out specific details of how things are composed.  That is more the purpose of developer documentation, imo.

These things can and will change over time in Decred due to the ability for consensus rule voting.  It is almost positively certain this glossary will become outdated and incorrect if it includes specific details.

The glossary should, imo, describe the **purpose** of a coinbase transation, not how it actually fulfills that purpose.  The same applies to other concepts as well.

docs/glossary.md
```diff
@@ -115,11 +131,28 @@ Tickets that reached the end of their window without being called to vote - thes
 
 A mechanism for obtaining free (testnet) coins.
 
+#### Full node
+
+An instance of the Decred daemon `dcrd` which is serving the full version of the blockchain to peers, and fully verifying all of the consensus rules of Decred.
```
davecgh: See the description of a full node in the dcrd README.  This is not accurate.  Full does not refer to whether or not it serves the full blockchain.

docs/glossary.md
```diff
@@ -115,11 +131,28 @@ Tickets that reached the end of their window without being called to vote - thes
...
+The very first block of the Decred blockchain, created on the 8th of February 2016.
+
 
 ## H
 
 ---
 
+#### Hardware wallet
+
+A dedicated hardware device which securely stores a wallets private keys. Examples include [Trezor](https://trezor.io/) and [Ledger](https://www.ledger.com/).
```
davecgh: Not all hardware wallets are necessarily all that secure.  I'm not sure that claim should be made here.

docs/glossary.md
```diff
@@ -184,6 +221,19 @@ A group of miners who share (pool) their computational resources to mine DCR. Wh
 
 Tickets that have been called but did not receive a reward. This can happen if a ticket is called to vote, but the wallet that bought the ticket does not respond. This can also happen if the wallet does respond and broadcasts its vote to the network, but a miner does not include their vote in the following block. 
 
+#### Multisignature
+
+Multisignature refers to Decred addresses which require more than one private key to authorize transactions. A multisignature address can support multiple keys (N) and a subset of those (M) are required to transact (commonly known as "MofN"). For example, an 2 of 3 address would have three valid keys, however only two of them would be required to authorize transactions.
```
davecgh: Mutlisignature does not refer to addresses, rather it refers to the conditions required to spend funds.

docs/glossary.md
```diff
@@ -184,6 +221,19 @@ A group of miners who share (pool) their computational resources to mine DCR. Wh
...
+Multisignature refers to Decred addresses which require more than one private key to authorize transactions. A multisignature address can support multiple keys (N) and a subset of those (M) are required to transact (commonly known as "MofN"). For example, an 2 of 3 address would have three valid keys, however only two of them would be required to authorize transactions. 
+
+
+## N
+
+---
+
+#### Nonce
+
+The word nonce is derived from "number used once". In the context of Decred, this usually refers to a 4-byte field in the block header which is adjusted by proof-of-work miners so that the hash of the block is lower than or equal to the current difficulty target of the network.
```
davecgh: > ... block header which, _along with some additional bytes used as an extra nonce_, is adjusted by proof-of-work miners ....

docs/glossary.md
```diff
@@ -290,6 +340,10 @@ A wallet mode in which only blocks related to addresses owned by the wallet are
 
 An interval of 2016 blocks (~1 week) which is used to determine if a vote on concensus rules can begin. Before a vote on consensus rule changes can begin, 75% of tickets that vote during a Stake Version Interval (SVI) must be using software that contains the latent software change being proposed.
 
+#### Stakebase transaction
+
+Every block which is mined will contain a stakebase transaction for each ticket which voted on that block. Stakebase transactions have two inputs, the Decred which was spent to purchase the ticket, and the newly created Decred constituting the reward for voting. Stakebase transactions are also created when an expired or missed ticket is revoked, however there will be no voting reward created in these cases. Stakebase transactions belong to the stake transaction tree, which means that these transactions cannot be rejected by proof-of-stake voters, even if they vote to reject the block which contains them.
```
davecgh: Again, the specific number of inputs and outputs is almost positively going to change over time.  I highly suggest to describe the purpose, not the specific technical details of the transaction structure.

jholdstock:
I understand the point @davecgh, but I think I agree with @s-ben here.

- If the consensus rules change then the documentation will fall out date regardless of where it lives, and regardless of whether we call it "user docs" or "dev docs". So personally I don't believe the "it will become outdated" argument carries much weight.
- I don't believe the specific details called out here (eg. coinbase transaction details) are "dev specific" knowledge. These details are useful for an everyday user who is browsing dcrdata or just trying to understand how their currency works. I personally am not a protocol or consensus developer, but I still found it useful to learn about these things.
- These details are better documented in arguably-the-wrong-place than not documented anywhere at all. We could always move the content later if a more appropriate place is created, but right now I don't think one exists.

docs/glossary.md
jholdstock: Is this more accurate?

> Multisignature refers to transactions which can be authorized by more than one private key. Multisignature can support multiple keys (N) and a subset of those (M) are required to transact (commonly known as "MofN"). For example, an 2 of 3 multisignature would have three valid keys, however only two of them would be required to authorize transactions. 

I may not have a deep enough understanding of this to produce an accurate definition without an existing definition to work from.

docs/glossary.md
jholdstock: Sounds good to me. I've added a link.

docs/glossary.md
jholdstock: Copied the description from the dcrd README

docs/glossary.md
jholdstock: Removed the word "secure"

docs/glossary.md
jholdstock: Extra wording included

docs/glossary.md
davecgh: I think just adding the word `transactions` after multisignature in the latter two cases would do the trick.  It would be accurate then.

jholdstock: Done. Thanks

jholdstock:
I have incorporated all of the suggestions from @davecgh, @s-ben and @teknico. I think this PR is good to  go

PR #801, opened 2019-01-23
Adds address details page

### LINKED ISSUE #710: Various address types Dc, Ds, Dk
Dave C mentioned various address types and their associated prefixes here: https://github.com/decred/dcrdocs/pull/675#discussion_r235489109

Might be a nice detail to mention somewhere.

### LINKED ISSUE #779: Create page on Addresses for advanced section
Recently, we've added a fair bit of technical information on addresses. 

- In this [PR](https://github.com/decred/dcrdocs/pull/731), we expanded the address definition in glossary: 
- In this [Issue](https://github.com/decred/dcrdocs/issues/710), we're collaborating on text describing various types of addresses, @zubairzia0 has added a great table documenting address types, and @jholdstock has linked to a well-written forum post by @davecgh that could be reused.  

This is all well and good, but it's too much information for a glossary definition (which is already bloated IMO). I propose creating a new page under /Advanced titled 'Addresses' (or 'Address Details') that collects all the information above. We could then also shorten the bloated definition in the glossary, and link to the new page from there for users that want more technical information.

s-ben:
This PR creates a 'Address details' page under /Advanced ( Closes #779 , Closes #710 ). Previously, technical details about addresses were found in the glossary definition for Address and a couple FAQ questions. An advanced page seemed like a better place to aggregate this information. Have also added the table @zubairzia0 added in Issue https://github.com/decred/dcrdocs/issues/710 (@jholdstock maybe this finally closes that issue?). The glossary definition of address and relevant FAQ questions have also been updated to be more concise and point to the new address page.

### DIFF
--- a/docs/advanced/address-details.md
+++ b/docs/advanced/address-details.md
@@ -8,11 +8,9 @@ The most commonly used address currently is the Pay-to-PubkeyHash (P2PKH) type, 
 
 ## Address format
 
-Each address has a 2-byte prefix than can be used to identify its type. The first byte of the prefix identifies the network. Mainnet addresses start with `D`, testnet addresses start with `T`, simnet addresses start with `S`, and regnet addresses start with `R`. The second byte identifies the address type. Pay-to-PubkeyHash (P2PKH) addresses, for example, contsin a lowercase `s` in the second byte. So, a P2PKH address on mainnet will begin with`Ds`, as shown in the sample address below.  
+Each address has a 2-byte prefix than can be used to identify its type, and a checksum suffix to detect improperly entered addresses. The first byte of the prefix identifies the network. Mainnet addresses start with `D`, testnet addresses start with `T`, simnet addresses start with `S`, and regnet addresses start with `R`. The second byte identifies the address type. Pay-to-PubkeyHash (P2PKH) addresses, for example, contsin a lowercase `s` in the second byte. So, a P2PKH address on mainnet will begin with`Ds`, as shown in the sample address below.  
 
-```DsYszZ5n79HJzyTVGAdXeDb1b3Z5N2Tnsij```
-
-
+```DsExampleAddr1For2Demo3PurposesOnly```
 
 The table below provides the locking script and prefix details for each address type. 
 
--- a/docs/advanced/address-details.md
+++ b/docs/advanced/address-details.md
@@ -1,0 +1,31 @@
+# <img class="dcr-icon" src="/img/dcr-icons/Transactions.svg" /> Address Details 
+
+---
+
+A Decred address is a representation of a public key, which itself could be a script hash. 
+
+The most commonly used address currently is the Pay-to-PubkeyHash (P2PKH) type, which is used to send DCR to a single address. For this address type, the recipient must provide both the public key which hashes to the requisite hash, and a valid signature that is created from the corresponding private key. Other address types require signatures from multiple parties (multi-sig), signatures created within specified time limits, or other more complex constraints. Pay-to-Script-Hash (P2SH) addresses, for example, generate a 1-of-2 multisig transaction utilized by Voting Service Providers (VSPs) that allow VSPs to vote on behalf of ticket holders. In this case, both the ticket holder and VSP have their own private keys, allowing either to vote. 
+
+## Address format
+
+Each address has a 2-byte prefix than can be used to identify its type. The first byte of the prefix identifies the network. Mainnet addresses start with `D`, testnet addresses start with `T`, simnet addresses start with `S`, and regnet addresses start with `R`. The second byte identifies the address type. Pay-to-PubkeyHash (P2PKH) addresses, for example, contsin a lowercase `s` in the second byte. So, a P2PKH address on mainnet will begin with`Ds`, as shown in the sample address below.  
+
+```DsYszZ5n79HJzyTVGAdXeDb1b3Z5N2Tnsij```
+
+
+
+The table below provides the locking script and prefix details for each address type. 
+
+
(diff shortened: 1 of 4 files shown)


docs/faq/wallets-and-seeds.md
```diff
@@ -70,27 +70,7 @@ A public key address, also called Pay-To-Pubkey (P2Pk), can be identified with i
 
 #### 8. What are the different types of addresses?
 
-A Decred address[^14995] is actually just a representation of a public key (which itself could be a script hash) along with a 2-byte prefix which identifies the network and type and a checksum suffix in order to detect improperly entered addresses.
```
jholdstock: > a checksum suffix in order to detect improperly entered addresses.

The new page loses this detail. Can we include it please? Maybe including how many bytes the checksum consists of.

jholdstock: As we are dropping this line, we no longer need the reference [^14995] at the bottom of this file. We can delete that line as well

docs/faq/wallets-and-seeds.md
```diff
@@ -70,27 +70,7 @@ A public key address, also called Pay-To-Pubkey (P2Pk), can be identified with i
 
 #### 8. What are the different types of addresses?
```
jholdstock: Seeing as we now have both the glossary entry and the whole page dedicated to addresses, do you think we could just drop this FAQ entry altogether? To me it feels like we have enough content on this subject and we dont need the FAQ as well

docs/faq/wallets-and-seeds.md
```diff
@@ -70,27 +70,7 @@ A public key address, also called Pay-To-Pubkey (P2Pk), can be identified with i
...
-
-Voting Service Providers (VSPs) use pay-to-script-hash addresses, which are identified by the second byte being a lowercase "c" (again that is shown in the linked params). The specific flavor of script it generates is a multi-signature 1-of-2, which is how it allows either the VSP, or you, to vote. Both you and the VSP have your own private keys and since the script only requires one signature of the possible two, that is how it allows delegation of voting rights to the VSP without you giving up your voting rights completely.
-
-| Address   Type     	| Locking   Script 	| (Decred) Mainnet  	| Testnet 	| Simnet 	| Prefix Size (byte) 	|
-|--------------------	|:----------------:	|:-----------------:	|:--------:	|:------:	|:------------------:	|
-| Pay-to-Pubkey      	|       P2Pk       	|         Dk        	|    Tk   	|   Sk   	|    2                  	|
-| Pay-to-Pubkey-Hash (secp256k1)  	|       P2PKH      	|         Ds        	|    Ts   	|   Ss   	|    2                  	|
-| Pay-to-Script-Hash 	|       P2SH       	|         Dc        	|    Tc   	|   Sc   	|    2                  	|
-
----
```
jholdstock: If we decide to keep this FAQ entry, this page break should be re-instated. If the FAQ is removed, the page break can go too.

docs/advanced/address-details.md
```diff
@@ -0,0 +1,31 @@
...
+
+## Address format
+
+Each address has a 2-byte prefix than can be used to identify its type. The first byte of the prefix identifies the network. Mainnet addresses start with `D`, testnet addresses start with `T`, simnet addresses start with `S`, and regnet addresses start with `R`. The second byte identifies the address type. Pay-to-PubkeyHash (P2PKH) addresses, for example, contsin a lowercase `s` in the second byte. So, a P2PKH address on mainnet will begin with`Ds`, as shown in the sample address below.  
+
+```DsYszZ5n79HJzyTVGAdXeDb1b3Z5N2Tnsij```
```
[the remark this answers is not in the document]
s-ben: Changed address to example provided above,
``````DsExampleAddr1For2Demo3PurposesOnly```

A testnet might look better, but couldn't find one easily. If someone posts one here can use that instead.

docs/faq/wallets-and-seeds.md
s-ben: Good catch. Was meaning to include that, but got lost in my wordsmithing:)

Have added detail to first para:

> Each address has a 2-byte prefix than can be used to identify its type, and a checksum suffix to detect improperly entered addresses. The first byte of the prefix identifies the network...

docs/faq/wallets-and-seeds.md
s-ben: If this was only in a page under /Advanced, I'd say keep it, as it's not easily discoverable. But with the address definition in the glossary (which links to this page), I think we should remove this (have removed in last push, but can add back if someone strongly disagrees).

s-ben:
@jholdstock have addressed all comments. I do think a more detailed reference like the [bitcoin address page](https://en.bitcoinwiki.org/wiki/Technical_background_of_version_1_Bitcoin_addresses) you point out would be good. I know @davecgh likes them diagrams:) Would be happy to at least update the Bitcoin diagram for Decred-specific deviations.
image

docs/faq/wallets-and-seeds.md
s-ben: I can also add the number of bytes the checksum uses if someone can provide. I assume it's 4 like Bitcoin addresses, but don't want to assume and can't find this in our existing docs.

davecgh: 2-byte prefix, fingerprint, 4-byte checksum.  The checksum is 4 bytes of `blake256(blake256(prefix||checksum))`.

docs/advanced/address-details.md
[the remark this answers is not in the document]
jholdstock: Here is a valid testnet address `TsSpRM32SQLkWQk6WPkB1p8PcrCujdSkf5C`

PR #810, opened 2019-01-24
Remove PoW reward from block 1.

### LINKED ISSUE #628: Validate block 1 details on the inflation page
https://docs.decred.org/advanced/inflation/

Was the block reward for block 1 actually 1,680,000?
Was this actually rewarded to PoW miners?

jholdstock:
Stealing DaveC's words from chat :)

Closes #628

### DIFF
--- a/docs/advanced/inflation.md
+++ b/docs/advanced/inflation.md
@@ -18,11 +18,11 @@ The following chart shows an estimate of the coin supply growth over time.
 
 ![Decred supply chart](/img/decred_supply.png)
 
-The table below shows the estimated block reward and estimated total supply of Decred up to block 2,457,600 in 2039. Note that the total supply of DCR at block 1 is 1,680,000 due to the [premine](../faq/general.md#3-how-was-the-decred-premine-distributed). PoS voting started at block 4,096[^3] therefore PoS rewards were not generated before that height.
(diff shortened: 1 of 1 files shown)


docs/advanced/inflation.md
```diff
@@ -18,11 +18,11 @@ The following chart shows an estimate of the coin supply growth over time.
...
+The table below shows the estimated block reward and estimated total supply of Decred up to block 2,457,600 in 2039. Note that the total supply of DCR at block 1 is 1,680,000 due to the [premine](../faq/general.md#3-how-was-the-decred-premine-distributed). There was no block reward for proof-of-work miners in block 1, which allowed miners to voluntarily choose whether or not to accept the terms of the airdrop independent of a potential reward which could skew the incentive. PoS voting started at block 4,096[^3] therefore PoS rewards were not generated before that height.
 
 Block height | Estimated date     | Block reward (DCR) | PoW (DCR)    | PoS vote (DCR) | Decred Treasury (DCR) | Total DCR supply
 ------------ | ------------------ | ------------------ | ------------ | -------------- | ----------------- | ----------------
-1            | February 8, 2016   | 1,680,000.00       | 1,680,000.00 | 0.00           | 0.00              | 1,680,000
+1            | February 8, 2016   | 1,680,000.00       | 0.00         | 0.00           | 0.00              | 1,680,000
```
davecgh: I don't think this should be removed.  It was, in fact, generated by PoW.  I suppose it might just be a case of how you choose to read it, but I personally think it should be included to indicate it was generated by PoW miners.

jholdstock: Do you not think the paragraph above clarifies that point? If the first row is ignored, the numbers in this column indicate how much DCR was awarded to the proof-of-work miner who produced the block. Row 1 is an outlier.

PR #820, opened 2019-01-29
Host in nginx.

jholdstock:
- Host in nginx instead of httpd
- Move `.sh` scripts into `bin/`
- Also upgrade to mkdocs 3.3.0.

Good to go.


jholdstock:
Rebased onto master. Takes `nginx.conf` from dcrweb. Also includes upgrading to mkdocs 3.3.0.

Good to go.

PR #825, opened 2019-02-05
Jargon change cleanup

s-ben:
While we changed the 'Solo Mining' page to 'Solo Voting' under the PoS FAQ, the menu is still displaying 'Solo Mining'. Changed to 'Solo Voting' to match page title.


s-ben:
Done. @jholdstock you may want to double check the nginx redirect though? It looks correct, but I tried the redirect locally and it didn't work. Presumably this is the expected behavior because redirects only work in prod, but just wanted to make sure.

jholdstock:
Thanks, looks good to me. Unfortunately yes - due to a limitation in how nginx does redirection, the redirects only work in production and not locally

PR #838, opened 2019-02-08
Adding P2PKH gen process flow

### LINKED ISSUE #806: Diagram showing how addresses are composed
In #801 we discussed adding a diagram to the docs which shows how addresses are composed, similar to the following.

In an ideal world we would not simply include this as an image, but instead include some kind of source file which means that the chart can be easily updated. 

51638521-63c1dd80-1f13-11e9-854b-b459d88db24c

zubairzia0:
Added a diagram (SVG file) that shows how P2PKH addresses are generated.
(Closes #806)


zubairzia0:
the sanitize portion was something I put because there were talks about SVG potentially having malicious code. I do see it render properly on my end though. However, I think it's possible that it messes up some formatting of the image so I have changed the html to markdown syntax. Also removed centering of the image.

xaur:
Small nit: do not abuse, or don't use at all CAPS in file names, also don't use spaces.

`?sanitize=true` is not relevant here. I guess it is a command to a server that serves the file to process it. If dcrdocs don't recognize such command, nothing happens.

zubairzia0:
abiding by lower cases with underscores.

jholdstock:
Hi @zubairzia0, are you able to correct the font display issue I described above?

One option could be to display a rendered .png file in the website, and we could store the source .svg image in a new directory in the repo root, maybe `img_src/`

zubairzia0:
@jholdstock , I inserted using mk syntax and removed the html that had the sanitize component.  The image appears as it does in inkscape.

Internet explorer
image

chrome:
image

jholdstock:
I'm going to merge this for now and create a follow-up PR with some extra changes to fix apparent browser specific display issues.

PR #841, opened 2019-02-08
Reorganizes FAQs, renames 'What is Proof-of-Stake?' page 

### LINKED ISSUE #814: No way to link to category
Someone [asked](https://www.reddit.com/r/decred/comments/ajpi3s/voting_service_providers_tech_questions/) a question about where to start and I wanted to share one "root" link for PoS that would overview and link to all "nested" pages. While there is a "Proof of Stake" item in the sidebar tree, there is no page corresponding to it, only nested pages. i.e. there's no

https://docs.decred.org/proof-of-stake/

I also wanted to link to "the FAQs" but again there's no page for "PoS FAQ", i.e. here:

https://docs.decred.org/faq/proof-of-stake/

### LINKED ISSUE #833: Move all FAQ pages under the top-level FAQ menu item
[As discussed here](https://github.com/decred/dcrdocs/issues/814)

s-ben:
Moves the PoS FAQ pages to under the top-level FAQ menu ( Closes #833 ). 

Renames the 'What is Proof-of-Stake?' page to 'Overview', as discussed [here](https://github.com/decred/dcrdocs/issues/814), making it consistent with the PoW 'Overview' page.

Also, while I was here, did some minor cleanup around naming consistency. Just changing instances of 'Proof of Stake/Work' with 'Proof-of-Stake (PoS'), Proof-of-Work (PoW), to make them consistent with other pages in the docs. 

@jholdstock , realize I'm bundling together a few small issues here. Can break into separate PRs if you think that's better.

### DIFF
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -81,14 +81,8 @@ nav:
       - 'dcrctl RPC Commands': 'wallets/cli/dcrctl-rpc-commands.md'
     - 'SPV': 'wallets/spv.md'
 - Proof-of-Stake Voting:
-  - "What is Proof-of-Stake?": 'proof-of-stake/proof-of-stake.md'
+  - "Overview": 'proof-of-stake/proof-of-stake.md'
   - "How to Stake": 'proof-of-stake/how-to-stake.md' 
-  - PoS FAQ:
-    - 'General': 'faq/proof-of-stake/general.md'
-    - 'Solo Voting': 'faq/proof-of-stake/solo-voting.md'
-    - 'Buying Tickets and Fees': 'faq/proof-of-stake/buying-tickets-and-fees.md'
-    - 'Voting Tickets': 'faq/proof-of-stake/voting-tickets.md'
-    - 'Voting Service Providers': 'faq/proof-of-stake/voting-service-providers.md'
 - Proof-of-Work Mining:
   - 'Overview': 'mining/proof-of-work.md'
   - 'gominer Pool Mining': 'mining/proof-of-work/pool-mining/gominer.md'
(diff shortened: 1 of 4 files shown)


xaur:
I think this shouldn't close #814 until a decent solution is found or at least a workaround is applied widely. The latter means a) adding missing Overview/Intro pages, b) making sure they can serve as entry points - guiding people through the rest of the section. If it requires to spawn a dozen Overview/Intro pages where they don't fit that's not a nice solution too, but in general I believe having such entry points is useful.

s-ben:
@xaur , edited PR so it won't close #814.

dajohi:
rebase needed.

s-ben:
Ran into conflict while rebasing, but should be good to go (everything looks correct when I run locally).

mkdocs.yml
```diff
@@ -99,6 +93,13 @@ nav:
   - 'Wallets and Seeds': 'faq/wallets-and-seeds.md'
   - 'Blocks': 'faq/blocks.md'
   - 'Common Errors and Solutions': 'faq/errors.md'
+  - 'PoW': 'faq/proof-of-work-mining.md'
```
jholdstock: This needs to be removed from its original position - it still exists under the "Proof-of-Work Mining" top level header.

s-ben: Oops. Fixed

xaur:
> changing instances of 'Proof of Stake/Work' with 'Proof-of-Stake (PoS'), Proof-of-Work (PoW)

Small question, why hyphens are needed?

jholdstock:
I don't think we have ever discussed whether the terms should be hyphenated or not, but the fact of the matter is that they are hyphenated 99% of the time on the docs site, so we should maintain that consistency.

PR #851, opened 2019-02-13
Resolve #794 by replacing stakepool with vsp, where possible

### LINKED ISSUE #794: Remove "stakepool" from Decrediton pages
#581 removed instances of "stakepool" in favour of VSP.

There are some outstanding instances of stakepool in the Decrediton pages because Decrediton still uses that terminology. Once Decrediton is updated, the docs can be updated.

isuldor:
This attempts to address issue #794. References to `dcrstakepool` have been left intact since it has not yet been renamed.


jholdstock:
Hi @isuldor,

Updating the links is a good idea, however I think we should keep the Decrediton pages as they are for now because Decrediton is still using the old terminology.

Hopefully we can get this sorted in Decrediton 1.5

isuldor:
@jholdstock understood, I've updated the PR accordingly

PR #859, opened 2019-02-16
Adds 'Upgrading Decrediton' page

### LINKED ISSUE #853: Create new page on upgrading Decrediton
Per this [discussion](https://matrix.to/#/!tfqymymiNgzSUJTHqS:decred.org/$155009444910322kWCcw:decred.org?via=decred.org&via=matrix.org&via=zettaport.com), we are getting people asking how to update Decrediton on Telegram. While the steps are simple, it appears documentation would be useful here. It also presents an opportunity to point users to important things to do after upgrading, such as look for new agendas. 

A logical place for this page seems to be Wallets / Decredition (GUI) / Upgrading Decrediton

s-ben:
This PR creates a new page with instructions on how to upgrade Decrediton ( Closes #853 ).


docs/wallets/decrediton/upgrading-decrediton.md
```diff
@@ -0,0 +1,16 @@
+# Upgrading Decrediton
+
+To upgrade Decrediton, simply download the new version and install it. You do not need to delete your existing version or perform any additional steps. When you open the upgraded version, your existing wallets and accounts will appear. 
+
+## Upgrading steps
+
+1. Download latest version. The latest downloads can be found at [https://decred.org/downloads/](https://decred.org/downloads/)
+or by clicking on the `Update Available` button in the upper right-hand corner of Decrediton.
+    <br>
```
jholdstock: Inline html shouldnt be required here, an empty new-line should do the trick.

docs/wallets/decrediton/upgrading-decrediton.md
```diff
@@ -0,0 +1,16 @@
+# Upgrading Decrediton
+
```
jholdstock: Could you insert a page-break here `---` as per the other pages.

docs/wallets/decrediton/upgrading-decrediton.md
```diff
@@ -0,0 +1,16 @@
...
+## Upgrading steps
+
+1. Download latest version. The latest downloads can be found at [https://decred.org/downloads/](https://decred.org/downloads/)
+or by clicking on the `Update Available` button in the upper right-hand corner of Decrediton.
+    <br>
+    ![Decrediton update link](/img/decrediton/upgrading/update-available.png)
+    Click on the link for your OS and download the installer.
+    ![Decrediton update link](/img/decrediton/upgrading/download-link.png)
+2. Close your current version of Decrediton (if open) and click on the downloaded installer. Follow the installation instructions as you would for any other program.
```
jholdstock: Remember to only use `1.` for every item of a numbered list, and then markdown will increment the numbers automatically.

jholdstock: > Follow the installation instructions as you would for any other program. 

Shall we add a link to `decrediton-setup.md` here?

docs/wallets/decrediton/upgrading-decrediton.md
s-ben: Forgot to remove that. I tried the line and the `<br>` tag actually, and neither introduced the space I was looking for between the image and the text above it. Looks OK anyway,so just removed the `<br>` (which wasn't doing anything).

docs/wallets/decrediton/upgrading-decrediton.md
s-ben: Done

docs/wallets/decrediton/upgrading-decrediton.md
s-ben: Right. Good best practice. Done.

s-ben: Since this page is for people upgrading Decrediton, I think this might be confusing. Most people have already created and configured their wallets using degredition-setup.md, and don't have a need to go back to it. I upgraded to 1.4.0 while making this page, and all I had to do was click through the usual windows to install the upgrade.

PR #869, opened 2019-03-05
Updates SPV disadvantages

### LINKED ISSUE #832: Update SPV disadvantages
Should mention that SPV wallets are not able to determine if a ticket has been missed, and so it is not possible to revoke a missed ticket until it has expired.

s-ben:
Updates SPV page to mention that SPV wallets are not able to determine if a ticket has been missed, and so it is not possible to revoke a missed ticket until it has expired ( Closes #832 )


docs/wallets/spv.md
```diff
@@ -36,7 +36,8 @@ These concerns do not apply to SPV wallets because they connect directly to the
 
 ## Are there any disadvantages?
 
-- SPV does not support voting wallets. Voting wallets have the responsibility to vote on the validity of the last block, and a wallet cannot be sure of the validity unless it fully validates the whole blockchain leading up to that block. It is possible however to purchase tickets and allocate the voting rights to a [Voting Service Provider](../proof-of-stake/how-to-stake.md#pos-using-a-voting-service-provider-vsp).
+- SPV does not support voting wallets. Voting wallets have the responsibility to vote on the validity of the last block, and a wallet cannot be sure of the validity unless it fully validates the whole blockchain leading up to that block. It is possible to purchase tickets and allocate the voting rights to a [Voting Service Provider](../proof-of-stake/how-to-stake.md#pos-using-a-voting-service-provider-vsp). However, because SPV wallets are not able to determine if a ticket has been missed, it is not possible to revoke a missed ticket until it has expired.
```
jrick: in practice, this isn't an issue for most spv users because the vsp will revoke missed tickets, and they are financially incentivized too in the current model because they receive payment for the revocation (but less payment than if a vote was mined).

s-ben: @jholdstock perhaps has better wording? Was there a particular incident that inspired https://github.com/decred/dcrdocs/issues/832 ?

jrick: "Be aware that if a VSP fails to revoke a missed ticket, ..."

s-ben: How about this?

> Be aware that in the very rare case of a VSP failing to revoke a missed ticket, it is not possible for an SPV wallet to revoke the missed ticket until it has expired.

jrick: Predicting a VSP's uptime in the official docs doesn't seem like a wise idea.

And technically a missed ticket can never expire (as it is no longer live), but a SPV wallet won't allow you to revoke it until it has reached the confirmation depth of an expired ticket.

s-ben: Took another stab:

> Be aware that if a VSP fails to revoke a missed ticket, it is not possible for an SPV wallet to revoke the missed ticket until the it has reached the confirmation depth of an expired ticket.

PR #897, opened 2019-03-28
Adds warning to back up redeem script for VSP

### LINKED ISSUE #884: Mention importance of backing up redeem script for VSP
At the moment we don't mention this in `dcrwallet-tickets.md` or anywhere else I can find

s-ben:
Tells users to save a copy of their multi-sig voting script when signing up for a VSP ( Closes #884 ). 

@jholdstock this was the only obvious place I could think to add this warning. Can you think of anywhere else?


jholdstock:
We probably want to add something in the decrediton guides for non-cli users.
Here perhaps? https://docs.decred.org/wallets/decrediton/using-decrediton/#tickets

xaur:
How about making this more visible, e.g. in a standalone yellow Notice block (if there is such a thing in mkdocs)?

s-ben:
@jholdstock have updated the Using Decrediton page as well.

@xaur have added warning text formatting to both places we're adding the backup warning/instructions. This is what it looks like on the Decrediton page,

image

PR #909, opened 2019-04-04
Fix broken redirects.

jholdstock:


noahpierau:
Was about to open an issue for the https://docs.decred.org/proof-of-stake/proof-of-stake/ page, but it seems you already have it covered. Were these the only broken redirects?

jholdstock:
Yep just those two. Its a pain that we cannot test these locally any more due to the way nginx handles redirects. We need to wait until deployment to production before we are able to validate.

PR #924, opened 2019-04-18
Adding decred canon to beginner-guide.md

zubairzia0:


s-ben:
One minor nit. The 'A' is capitalized in the description: `A collection of readings and resources on Decred.`, whereas the other items in the list do not capitalize the first word. However, the other lists on the page mix capitalization too... Can just fix myself in another PR if nobody else does it.

Otherwise looks good.

Issue #615, opened 2018-09-18
Bringing some FAQ/Advanced gems to the fore


RichardRed0x:
The information in FAQ, Research and Advanced sections is sometimes not well labelled or duplicative of other sections. There are some pages or snippets that are worth bringing more to the fore through either a link or embedding some info in other higher-level pages. This issue is for collecting a list of relatively obscure docs content that should be made more visible.

[https://docs.decred.org/faq/general/#bring-up-costs-840000-coins-50-of-premine-4-of-total-decred-supply](https://docs.decred.org/faq/general/#bring-up-costs-840000-coins-50-of-premine-4-of-total-decred-supply) - nice description of the pre-mine and airdrop.  This could be linked to from a page like governance, but it might be worth starting a new page like Origins or History, with stuff like this and the C0 blog posts and maybe some information about past consensus rules changes, all in a nice accessible package. 

The [Hybrid Design](https://docs.decred.org/research/hybrid-design/) page in Research section has a good description of the mechanics of "block voting", and the only place I know where the scenario of voting No on a regular transaction tree is discussed. This could perhaps be integrated with some other information about block voting to put all of that in a single page.

xaur:
Premine is often discussed, it is important historic event and fully deserves own page. For people working on this, merge any missed info from [this chat](https://matrix.to/#/!lbzTjhzNbIaDbuAxkS:decred.org/$15540621201905YGdoC:decred.org).

> it might be worth starting a new page like Origins or History, with stuff like this and the C0 blog posts and maybe some information about past consensus rules changes, all in a nice accessible package.

Relevant WIP: https://github.com/artikozel/decred-timeline

Past consensus rules pages have their own page now: https://docs.decred.org/governance/consensus-rule-voting/consensus-vote-archive/

Overall I think FAQ pages should carry minimum content and strive "dispatch" to real pages.

RichardRed0x:
I like the idea of a page about the airdrop/premine and genesis block - could be extended to a more complete history of significant events. Later events might be better on their own page, depending on how detailed the genesis history page gets.

PR #926, opened 2019-04-19
Remove redundant questions PoS FAQ

s-ben:
As noted in https://github.com/decred/dcrdocs/issues/615, some of our FAQ Qs have collected info that has become redundant, or not typical FAQ material (e.g. big topics better handled in a "main" page). This commit does some pruning of the PoS FAQ section. Specifically, it removes the following questions, which are already covered by the [PoS overview](https://docs.decred.org/proof-of-stake/overview/) page: 

- What are tickets?
- What happens to my funds when I buy a ticket?
- What are fees?
- What are the chances of my ticket voting?
- What is Proof-of-Stake (PoS) voting

There were a few small details in these answers that weren't in the PoS overview page. For these, I've incorporated those details into the overview page. 

We could probably remove more, or edit down existing answers to be more "bite sized", but I figure it's best to prune the low-hanging fruit first and keep the commit small.


s-ben:
P.S. accidentally created this branch from another branch I was reviewing. Had to manually remove other changes and force push. Can redo PR from scratch if need be.

docs/faq/proof-of-stake/general.md
```diff
@@ -2,21 +2,15 @@
 
 ---
 
-#### 1. What are tickets?
-
-A ticket is a token you buy to participate in the PoS system. You can buy tickets through the command line interface of dcrctl, or through a GUI like Decrediton[^1]. Whenever you buy a ticket, you pay the current ticket price and a ticket fee using your DCR. When a ticket is purchased, it goes to a temporary "mempool". Twenty tickets can be mined into each block and tickets are chosen to be mined based on their fee per kilobyte. Once your ticket is mined, it will move from the current "allmempool" to the "immature" ticket pool. After 256 blocks (about 20 hours) it then will mature and go into the live ticket pool where it is eligible to be chosen to vote.
```
jholdstock: You removed the link to reference `[^1]` here, which means the reference can be removed from the foot of the page.

docs/faq/proof-of-stake/general.md
```diff
@@ -26,25 +20,7 @@ The ticket price is always refunded, no matter if your ticket votes, misses or e
...
-
-The PoS system utilizes two types of fees, a "txfee" (also known as a 'split' fee) and a "ticketfee".
-
-The txfee is a fee you pay the network for handling the transaction to buy your ticket. This fee is at default set to a low amount (0.0001 DCR/kB) and should not be changed.
-The ticketfee is the fee you pay if your ticket gets mined. This fee is an incentive for PoW miners to mine your ticket and add it to the ticket pool.
-
-In case the buying window is full, meaning there are more tickets being bought than there are available to be mined during that window, the ticketfee will be used to determine which tickets get put into
-the ticket pool. The tickets with the highest ticketfee will be selected by the PoW miners.
-
-Fees are calculated in DCR per kilobyte. Since a ticket is only about 300 bytes (for solo) or 540 bytes (for VSP) in size, the actual fees you pay will be lower than what you set as the specific fee amount.
```
jholdstock: We are losing the 540 bytes figure in this PR. Can we include it on the overview page where we mention that a solo purchase is 300 bytes?

docs/faq/proof-of-stake/general.md
s-ben: Removed

docs/faq/proof-of-stake/general.md
s-ben: Good catch. When incorporating the 540 bytes figure though, did realize that the math was a bit fuzzy on our current example on the overview page. We say that the 300 Bytes solo transaction will cost 0.1 DCR, but 0.3 DCR/kB X 300 Bytes = 0.09 DCR (assuming 1,000 bytes/kilobyte, not 1,024). Likewise 540 DCR using this equation would give a cost of 0.162 DCR. Assuming my math is correct, have updated text to:

> Therefore, with a higher transaction size, you will end up paying a higher absolute fee. For example, solo-staking ticket purchases consume around 300 Bytes, whereas staking with a VSP will consume around 540 Bytes. Therefore, with a a **Ticket Fee** of 0.3 DCR/kB, a solo-staked ticket would cost ~0.09 DCR, and a VSP-staked ticket would cost ~0.16 DCR. 

Also, removed the following sentence about the fee being refunded if the ticket isn't mined. It was a bit clunky, and we also provide that info in the Ticket Lifecycle section.

jholdstock: We discussed on Matrix to remove the 0.3 altogether and just refer to the actual fee of 0.0001

s-ben: [Discussed](https://matrix.to/#/!dhHYPTtCtvPSUfTepT:decred.org/$155604517624183TYTTU:decred.org?via=decred.org&via=matrix.org&via=zettaport.com) this a bit more in #support just to clarify for myself. This is the text we came up with:

>The **Ticket Fee** defaults to the minimum (0.0001 DCR/kB), which is typically sufficient.

jholdstock: I like the new text, but just want to highlight that we are now losing the estimated tx size completely. 

I dont think the 300/540 bytes figures are included anywhere else in the docs, so it might be nice to keep them somewhere.

s-ben: So, I was able to find these numbers on the [buying tickets](https://docs.decred.org/wallets/cli/dcrwallet-tickets/) page for the CLI Wallet. Since you can only do solo voting using the CLI wallet, I think this is OK. I was trying to work this back in on this page, but couldn't think of a non-awkward place as the concepts of solo voting and VSP aren't introduced until the end of the page.

jholdstock: Sounds good, lets go with it then. Thanks for checking

PR #937, opened 2019-05-10
Adds transaction example 

### LINKED ISSUE #765: Add example transaction to Transaction Details page
It is desirable to have an example transaction on the [Transaction Details](https://docs.decred.org/advanced/transaction-details/) page, similar to the block header example on the [Block Header Specifications](https://docs.decred.org/advanced/block-header-specifications/) page. 

@davecgh has provided the below example transaction, broken into relevant fields,
```
01000000 // ver, sertype
---- Prefix Section Start ----
02 // numinputs
--- 1st input
4ff28b534361e383b5b149df4572feb1fbc659cb07067c3cb6c684452ae79f54 // reversed prevout tx hash
01000000 // prevout idx
00 // prevout tree
ffffffff // seq
--- 2nd input
c619e6ade3f469b8bb60b8ae9b0599eeeca05cd2db251cadea443344eafdac57 // reversed prevout tx hash
01000000 // prevout idx
00 // prevout tree
ffffffff // seq
02 // num outputs
-- first output
91c2e26d02000000 // output amount
0000 // script version
19 // script length
76a914f2ccddc70f8b8bcea24a362bf404841d53cbebc088ac // script
-- second output
d676dd6801000000 // output amount
0000 // script version
19 // script length
76a9141a1f5c7e9f7696989dced5cb9b61464f95790b7288ac // script
00000000 // locktime
00000000 // expiry
---- Prefix Section End ----
---- Witness Section Start ----
02 // num witnesses
-- first witness
ca68865402000000 // input amount
a29c0400 // containing block height
04000000 // containing block index
6a // script length
4730440220139466bd10b1f071f9b4ac16ed434d
63d090613fd06470f7bca9bba2b6ea6a0a02203c
ad1297ab9384466b3f304cd61e7e4ffa9abc1ae9
fe1630207fae6f200da1680121024d540859b805
c5780f2f0da350df8757c1defb72e3381b252d20
874478a5549c // script
-- second witness
49743a8201000000 // input amount
9e9c0400 // containing block height
01000000 // containing block index
6b
483045022100da5b0fde58c4c57a88d96f83b02e
ddcb67b1b68ce95cc562c895f2e4e57bb4430220
1a8744aed654d2979bdd797187f6bf63ad1898c5
d5c585309eeca558742858b201210307e3d98b00
4b15561c0f5aa158c90f93de33756d3bfd1f8a10
58a4d90ffc7ec7 // script
---- Witness Section End ----
```

The fields still need to be fully converted, which can be done by looking transaction up on dcrdata:
https://explorer.dcrdata.org/api/tx/decoded/a58389bd42520a9b9ecbae5432e27b91fde74d2f8e74ac6f8c22cda57b8ae348?indent=true

Ideally, this would be in a table.

s-ben:
Adds an example transaction to the [Transaction Details](https://docs.decred.org/advanced/transaction-details/) page ( Closes #765 ). 

Note: the transaction fields were broken down according to the example tx provided in #765 by @davecgh, and when parsing the data I ended up moving a couple bytes around. E.g. `script version` in the example was `0000`, but the docs say this is a two byte field, so have used `00` and moved the subsequent `00` into the next field (`script length`). There are also a couple fields that need double checking, such as `public key signature length` (hex evaluating to 25, though looks like the script has 50 bytes?) and `signature script length`. 

Also, the markdown table is a bit wonky, due to the fact that markdown doesn't support spanning cells. Can tweak the formatting with inline html, but wanted to get a first pass review on the current table first. We could also just use text formatting like we do for the block header [example](https://docs.decred.org/advanced/block-header-specifications/).

Also, do we want to paste a link to the full decoded tx JSON?
https://explorer.dcrdata.org/api/tx/decoded/a58389bd42520a9b9ecbae5432e27b91fde74d2f8e74ac6f8c22cda57b8ae348?indent=true

### DIFF
--- a/docs/advanced/transaction-details.md
+++ b/docs/advanced/transaction-details.md
@@ -132,7 +132,7 @@ Below is a breakdown of the above raw hex transaction data into fields. This tra
 <tr>
 <td>Tree</td>
 <td><code>00</code></td>
-<td>0 <br>(use regular transaction fee)</td>
+<td>0 <br>(regular transaction tree)</td>
 </tr>
 <tr>
 <td>Sequence number</td>
@@ -155,7 +155,7 @@ Below is a breakdown of the above raw hex transaction data into fields. This tra
 <tr>
 <td>Tree</td>
 <td><code>00</code></td>
-<td>0 <br>(use regular transaction fee)</td>
+<td>0 <br>(regular transaction tree)</td>
 </tr>
 <tr>
 <td>Sequence number</td>
@@ -183,7 +183,7 @@ Below is a breakdown of the above raw hex transaction data into fields. This tra
 <tr>
 <td>Public key script</td>
 <td style="word-break: break-all;"><code>76a914f2ccddc70f8b8bcea24a362bf404841d53cbebc088ac</code></td>
-<td style="word-break;">OP_DUP OP_HASH160 f2ccddc70f8b8bcea24a362bf404841d53cbebc0 OP_EQUALVERIFY OP_CHECKSIG <br>(Pay-To-Pubkey (P2Pk) hash script)</td>
+<td style="word-break: break-all; word-wrap: break-word;">OP_DUP OP_HASH160 f2ccddc70f8b8bcea24a362bf404841d53cbebc0 OP_EQUALVERIFY OP_CHECKSIG <br><br>(Pay-To-Pubkey-Hash (P2PKH) script)</td>
 </tr>
 <tr>
 <td colspan="3" style="text-align: center;">2nd output</td>
@@ -207,7 +207,7 @@ Below is a breakdown of the above raw hex transaction data into fields. This tra
 <tr>
 <td>Public key script</td>
 <td style="word-break: break-all;"><code>76a9141a1f5c7e9f7696989dced5cb9b61464f95790b7288ac</code></td>
-<td style="word-break;">OP_DUP OP_HASH160 1a1f5c7e9f7696989dced5cb9b61464f95790b72 OP_EQUALVERIFY OP_CHECKSIG <br>(Pay-To-Pubkey (P2Pk) hash script)</td>
+<td style="word-break: break-all; word-wrap: break-word;">OP_DUP OP_HASH160 1a1f5c7e9f7696989dced5cb9b61464f95790b72 OP_EQUALVERIFY OP_CHECKSIG <br><br>(Pay-To-Pubkey-Hash (P2PKH) script)</td>
 </tr>
 <tr>
 <td>Lock time</td>
@@ -250,7 +250,8 @@ Below is a breakdown of the above raw hex transaction data into fields. This tra
 <tr>
 <td>Signature script</td>
 <td style="word-break: break-all;"><code>4730440220139466bd10b1f071f9b4ac16ed434d63d090613fd06470f7bca9bba2b6ea6a0a02203cad1297ab9384466b3f304cd61e7e4ffa9abc1ae9fe1630207fae6f200da1680121024d540859b805c5780f2f0da350df8757c1defb72e3381b252d20874478a5549c</code></td>
-<td>2 data pushes. 1st push is DER signature. 2nd push is public key</td>
+<td style="word-break: break-all;">
+  The signature script contains two <br>pieces of data to be pushed onto<br> the stack:<br><br>1st push<br>(DER signature (asm))<br><code>30440220139466bd10b1f071f9b4ac16ed434d63d090613fd06470f7bca9bba2b6ea6a0a02203cad1297ab9384466b3f304cd61e7e4ffa9abc1ae9fe1630207fae6f200da16801</code><br><br>2nd push<br>(public key)<br><br><code>024d540859b805c5780f2f0da350df8757c1defb72e3381b252d20874478a5549c</code></td>
 </tr>
 <tr>
 <td colspan="3" style="text-align: center;">2nd witness</td>
@@ -278,7 +279,7 @@ Below is a breakdown of the above raw hex transaction data into fields. This tra
 <tr>
 <td>Signature script</td>
 <td style="word-break: break-all;"><code>483045022100da5b0fde58c4c57a88d96f83b02eddcb67b1b68ce95cc562c895f2e4e57bb44302201a8744aed654d2979bdd797187f6bf63ad1898c5d5c585309eeca558742858b201210307e3d98b004b15561c0f5aa158c90f93de33756d3bfd1f8a1058a4d90ffc7ec7</code></td>
-<td>2 data pushes. 1st push is DER signature. 2nd push is public key</td>
+<td style="word-break: break-all;">The signature script contains two <br>pieces of data to be pushed onto<br> the stack:<br><br>1st push<br>(DER signature (asm))<br><code style="word-break: break-all">3045022100da5b0fde58c4c57a88d96f83b02eddcb67b1b68ce95cc562c895f2e4e 57bb44302201a8744aed654d2979bdd797187f6bf63ad1898c5d5c585309eeca558742858b201</code><br><br>2nd push<br>(public key)<br><br><code style="word-break: break-all;">0307e3d98b004b15561c0f5aa158c90f93de33756d3bfd1f8a1058a4d90ffc7ec7</code></td>
 </tr>
 </tbody>
 </table>
--- a/docs/advanced/transaction-details.md
+++ b/docs/advanced/transaction-details.md
@@ -75,3 +75,64 @@ The witness data of a transaction involves only its inputs. The included data fi
 * **1 (No witness)** - The transaction's prefix is the only data present.
 * **2 (Only witness)** - The transaction's witness data is the only data present. For each input, this includes its value, block height, block index, and signature script.
 
+## Example Transaction
+
+Below we examine an [example transaction](https://explorer.dcrdata.org/tx/a58389bd42520a9b9ecbae5432e27b91fde74d2f8e74ac6f8c22cda57b8ae348) from the Decred mainnet.
+
+### Raw Transaction (Hex)
+
+Here is the raw hex data for the example transaction.
+
+`01000000024ff28b534361e383b5b149df4572feb1fbc659cb07067c3cb6c684452ae79f540100000000ffffffffc619e6ade3f469b8bb60b8ae9b0599eeeca05cd2db251cadea443344eafdac570100000000ffffffff0291c2e26d0200000000001976a914f2ccddc70f8b8bcea24a362bf404841d53cbebc088acd676dd680100000000001976a9141a1f5c7e9f7696989dced5cb9b61464f95790b7288ac000000000000000002ca68865402000000a29c0400040000006a4730440220139466bd10b1f071f9b4ac16ed434d63d090613fd06470f7bca9bba2b6ea6a0a02203cad1297ab9384466b3f304cd61e7e4ffa9abc1ae9fe1630207fae6f200da1680121024d540859b805c5780f2f0da350df8757c1defb72e3381b252d20874478a5549c49743a82010000009e9c0400010000006b483045022100da5b0fde58c4c57a88d96f83b02eddcb67b1b68ce95cc562c895f2e4e57bb44302201a8744aed654d2979bdd797187f6bf63ad1898c5d5c585309eeca558742858b201210307e3d98b004b15561c0f5aa158c90f93de33756d3bfd1f8a1058a4d90ffc7ec7`
+
+
+### Transaction Breakdown
+
+Below is a breakdown of the above raw hex transaction data into fields. This transaction uses the full serialization format, so we can see witness data along with inputs and outputs. Note that hex values are in little-endian format.
+
+Field        | Value        | Description
+---          | ---          | ---
+Version      | `01`  | Version: 1
+Serialization format  | `00`  | Serialization: Full serialization
+Number inputs |  `000002`     | Inputs: 2                                              
+1st input | `4ff2,8b53,4361,e383,b5b1,49df,4572,feb1,fbc6,59cb,0706,7c3c,b6c6,8445,2ae7,9f54`                 | Transaction hash
+      | `0100,0000`                                              | Output index: 1
+    | `00`       | Tree: 0
+       | `ffff,ffff`                     | Sequence number: 4294967295
+2nd input | `c619,e6ad,e3f4,69b8,bb60,b8ae,9b05,99ee,eca0,5cd2,db25,1cad,ea44,3344,eafd,ac57`                 | Transaction hash
+      | `0100,0000`                                              | Output index: 1
+    | `00`       | Tree: 0
+       | `ffff,ffff`                     | Sequence number: 4294967295
+1st output | `91c2,e26d,0200,0000`                 | Output amount: 104.33
+      | `0000`                                              | Version: 0
+    | `19`       | Public key script (length): 25
+       | `76a914f2ccddc70f8b8bcea24a362bf404841d53cbebc088ac`                     | Public key script
+2nd output | `d676,dd68,0100,0000`                 | Output amount: 60.54
+      | `0000`                                              | Version: 0
+    | `19`       | Public key script (length): 25
+       | `76a9141a1f5c7e9f7696989dced5cb9b61464f95790b7288ac`                     | Public key script
+Lock time      | `00000000`  | Lock time: 0 
+Expiry     | `00000000`  | Expiry: 0
+Number witnesses     | `02`  | Witnesses: 2
+1st witness    | `ca68,8654,0200,0000`  | Input amount: 100.08
+    | `a29c,0400`  | Block height: 302242
+        | `0400,0000`  | Block index: 4
+        | `6a`  | Signature script (length): 106
+        | `4730440220139466bd10b1f071f9b4ac16ed434d63d090613fd06470f7bca9bba2b6ea6a0a02203cad1297ab9384466b3f304cd61e7e4ffa9abc1ae9fe1630207fae6f200da1680121024d540859b805c5780f2f0da350df8757c1defb72e3381b252d20874478a5549c`  | Signature script 
+2nd witness    | `4974,3a82,0100,0000`  | Input amount: 64.79
+    | `9e9c,0400`  | Block height: 302238
+        | `0100,0000`  | Block index: 1
+        | `6b`  | Signature script (length): 107
+        | `483045022100da5b0fde58c4c57a88d96f83b02eddcb67b1b68ce95cc562c895f2e4e57bb44302201a8744aed654d2979bdd797187f6bf63ad1898c5d5c585309eeca558742858b201210307e3d98b004b15561c0f5aa158c90f93de33756d3bfd1f8a1058a4d90ffc7ec7`  | Signature script 
+
+
+
+
+
+
+
+
+
+
+
+
--- a/docs/advanced/transaction-details.md
+++ b/docs/advanced/transaction-details.md
@@ -78,7 +78,7 @@ The witness data of a transaction involves only its inputs. The included data fi
 
 ## Example Transaction
 
-Below we examine an [example transaction](https://explorer.dcrdata.org/tx/a58389bd42520a9b9ecbae5432e27b91fde74d2f8e74ac6f8c22cda57b8ae348) from the Decred mainnet.
+Below we examine an example [transaction](https://explorer.dcrdata.org/tx/a58389bd42520a9b9ecbae5432e27b91fde74d2f8e74ac6f8c22cda57b8ae348) from the Decred mainnet.
 
 ### Raw Transaction (Hex)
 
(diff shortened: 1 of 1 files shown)


davecgh:
Each hex digit is a nibble (4 bits), so 2 bytes takes 4 hex digits.  The version is indeed `0000`.

s-ben:
So how is the verision = `1` in the decoded [JSON](https://explorer.dcrdata.org/api/tx/decoded/a58389bd42520a9b9ecbae5432e27b91fde74d2f8e74ac6f8c22cda57b8ae348?indent=true) from the explorer,
```
{
   "txid": "a58389bd42520a9b9ecbae5432e27b91fde74d2f8e74ac6f8c22cda57b8ae348",
   "version": 1,
```
I believe this was my reasoning/rationalization for my guesswork.

davecgh:
That's the transaction version.  The script version is a different version.

davecgh:
A couple of things:

* In the transaction format section, each script should be preceded with a script length field which is itself a variable number of bytes
* The values are all raw hex values and, as such, imo, shouldn't have additional commas which I find confusing

s-ben:
> In the transaction format section, each script should be preceded with a script length field which is itself a variable number of bytes

Added variable to table:
`Signature script length | The length of the signature script as a variable-length integer` 

> The values are all raw hex values and, as such, imo, shouldn't have additional commas which I find confusing

I found that confusing as well, but was just following the convention in the [block header example](https://docs.decred.org/advanced/block-header-specifications/). Do we want to get rid of the commas there too?

s-ben:
@davecgh with the changes made per above comment, I think we're close to a workable example. However, wondering if more can/should be documented here. Personally, I'd like to understand Decred's transactions at a lower level. E.g. how the variable length integers are parsed. I think this could be valuable to other developers as well. But I'm wondering what questions we're getting from developers specifically, and what are they trying to do generally. Presumably they need to be able to a) parse transactions and b) construct transactions. But what types of transactions? At what level of abstraction? Do we need to document stakebase and other types of transactions? What is Decred-specific, and what is already covered by the Bitcoin dev docs [transaction guide](https://bitcoin.org/en/transactions-guide#introduction)? Also, do we want to do a diagram, like this one from the Bitcoin dev guide (which is an open source SVG if we want to start with this)?

image

davecgh:
> I found that confusing as well, have removed. Was just following the convention in the [block header example](https://docs.decred.org/advanced/block-header-specifications/). Do we want to get rid of the commas there too?

Commas are acceptable between each byte (2 hex digits), but there was something strange about how they were formatted such that is was non-standard and confusing.

davecgh:
An SVG diagram would be helpful I think.  As far as the other questions, all transactions follow the format specified here.  Votes, Ticket Purchases, etc, are merely specific incantations of them.

s-ben:
I believe this (public domain) diagram from the Bitcoin wiki could be easily modified for our purposes,
https://en.bitcoin.it/wiki/File:TxBinaryMap.png

I am wondering if this text from the diagram is true for Decred as well?
> Scripts and DER encoding both use big-endian values, all other serializations use little-endian

Also, this may be totally enough for this page. However, when I pretend to be a developer looking at this, I still have questions. What are the rules for encoding different types of transactions? What is the byte map for specific types of scripts? Are they the same as Bitcoin, or are they different? Can I just use Bitcoin documentation for this? What types of scripts are supported? What is the byte map for Decred-specific transactions (e.g. transactions for buying, voting, and revoking tickets)? Do they follow some rules? Do I just need to look to the source code for answers to these questions?

davecgh:
The bit about script and DER encoding using big-endian values is correct.  Regarding the rest, that is way too much to put into a general transaction specification imo.  What you're discussing is separate sections which discuss the various types of *standard* transactions.

s-ben:
> The bit about script and DER encoding using big-endian values is correct.

Tried and failed to come up with a simple way to word this. Realized that this is already painstakingly documented for Bitcoin transactions, and that if we're not doing anything different (at this level of abstraction anyway), it's probably not worth re-documenting here.

> Regarding the rest, that is way too much to put into a general transaction specification imo. What you're discussing is separate sections which discuss the various types of standard transactions and scripts.

For this page, you're probably right. I do still think it's useful to provide more documentation/explanation of Decred-specific transactions (e.g. stakebase transactions). But that doesn't need to be this page.

Perhaps for this PR, we should just stop here and do a final round of review on what we have so far?
* Updating `Transaction Format` section with `Signature script length` variable
* Adding transaction example

davecgh:
I agree that it would be useful to document each one for sure, but I just don't think it belongs in the general transaction specification.  They are strictly subsets of generalized transactions.

davecgh:
In terms of review, the section at the bottom doesn't appear formatted correctly.

s-ben:
Agreed, we probably don't want to put any more detail in the general transaction specification. Have proposed documenting Decred-specific transactions further in an [Issue](https://github.com/decred/dcrdocs/issues/942) to gather feedback on the idea. 

Table formatting is fixed. Markdown formatting is limited (e.g. no merged cells) so had to convert to html to get something more readable.

davecgh:
Already discussed on matrix, but the table formatting is still not correct on some browsers such as Firefox.

s-ben:
Ok fixed table formatting for Firefox. Turns out this is a common problem, which gets into some thorny CSS issues. I ended up just hard-coding the text wrapping for cells with long strings; the other, more elegant solutions from SO weren't working. Is this hacky CSS OK? Have tested on Brave, Chrome, Firefox and Safari and rendering fine on all now.

davecgh:
Still looks broken to me in FF 66.0.5.

docs/advanced/transaction-details.md
```diff
@@ -75,3 +76,210 @@ The witness data of a transaction involves only its inputs. The included data fi
...
+</tr>
+<tr>
+<td>Lock time</td>
+<td><code>00000000</code></td>
+<td>0</td>
+</tr>
+<tr>
+<td>Expiry</td>
+<td><code>00000000</code></td>
+<td>0</td>
```
davecgh: The description is correct for the number, but it might be worth adding that 0 means there is no expiry.

docs/advanced/transaction-details.md
```diff
@@ -75,3 +76,210 @@ The witness data of a transaction involves only its inputs. The included data fi
...
+</tr>
+<tr>
+<td>Public key script</td>
+<td style="word-break: break-all;"><code>76a9141a1f5c7e9f7696989dced5cb9b61464f95790b7288ac</code></td>
+<td></td>
+</tr>
+<tr>
+<td>Lock time</td>
+<td><code>00000000</code></td>
+<td>0</td>
```
davecgh: The description is correct for the number, but it might be worth adding that 0 means there is no lock time.

docs/advanced/transaction-details.md
```diff
@@ -75,3 +76,210 @@ The witness data of a transaction involves only its inputs. The included data fi
...
+<tr>
+<td>Number witnesses</td>
+<td><code>02</code></td>
+<td>2</td>
+</tr>
+<tr>
+<td colspan="3" style="text-align: center;">1st witness</td>
+</tr>
+<tr>
+  <td>Input amount</td>
```
davecgh: Extra spaces here that are nowhere else.

docs/advanced/transaction-details.md
```diff
@@ -75,3 +76,210 @@ The witness data of a transaction involves only its inputs. The included data fi
...
+<tr>
+<td>Sequence number</td>
+<td><code>ffffffff</code></td>
+<td>4294967295</td>
+</tr>
+<tr>
+<td colspan="3" style="text-align: center;">2nd input</td>
+</tr>
+<tr>
+  <td>Transaction hash</td>
```
davecgh: Extra spaces not anywhere else.

docs/advanced/transaction-details.md
```diff
@@ -75,3 +76,210 @@ The witness data of a transaction involves only its inputs. The included data fi
...
+<td><code>ffffffff</code></td>
+<td>4294967295</td>
+</tr>
+<tr>
+<td colspan="3" style="text-align: center;">1st output</td>
+</tr>
+<tr>
+<td>Output amount</td>
+<td><code>91c2e26d02000000</code></td>
+<td>104.33</td>
```
davecgh: 104.33512081

docs/advanced/transaction-details.md
```diff
@@ -75,3 +76,210 @@ The witness data of a transaction involves only its inputs. The included data fi
...
+<td style="word-break: break-all;"><code>76a914f2ccddc70f8b8bcea24a362bf404841d53cbebc088ac</code></td>
+<td></td>
+</tr>
+<tr>
+<td colspan="3" style="text-align: center;">2nd output</td>
+</tr>
+<tr>
+<td>Output amount</td>
+<td><code>d676dd6801000000</code></td>
+<td>60.54</td>
```
davecgh: 60.54311638

docs/advanced/transaction-details.md
```diff
@@ -75,3 +76,210 @@ The witness data of a transaction involves only its inputs. The included data fi
...
+<td><code>02</code></td>
+<td>2</td>
+</tr>
+<tr>
+<td colspan="3" style="text-align: center;">1st witness</td>
+</tr>
+<tr>
+  <td>Input amount</td>
+<td><code>ca68865402000000</code></td>
+<td>100.08</td>
```
davecgh: 100.08029386

docs/advanced/transaction-details.md
```diff
@@ -75,3 +76,210 @@ The witness data of a transaction involves only its inputs. The included data fi
...
+<td style="word-break: break-all;"><code>4730440220139466bd10b1f071f9b4ac16ed434d63d090613fd06470f7bca9bba2b6ea6a0a02203cad1297ab9384466b3f304cd61e7e4ffa9abc1ae9fe1630207fae6f200da1680121024d540859b805c5780f2f0da350df8757c1defb72e3381b252d20874478a5549c</code></td>
+<td></td>
+</tr>
+<tr>
+<td colspan="3" style="text-align: center;">2nd witness</td>
+</tr>
+<tr>
+<td>Input amount</td>
+<td><code>49743a8201000000</code></td>
+<td>64.79</td>
```
davecgh: 64.79836233

docs/advanced/transaction-details.md
```diff
@@ -75,3 +76,210 @@ The witness data of a transaction involves only its inputs. The included data fi
...
+</tr>
+<tr>
+<td>Signature script length</td>
+<td><code>6a</code></td>
+<td>106</td>
+</tr>
+<tr>
+<td>Signature script</td>
+<td style="word-break: break-all;"><code>4730440220139466bd10b1f071f9b4ac16ed434d63d090613fd06470f7bca9bba2b6ea6a0a02203cad1297ab9384466b3f304cd61e7e4ffa9abc1ae9fe1630207fae6f200da1680121024d540859b805c5780f2f0da350df8757c1defb72e3381b252d20874478a5549c</code></td>
+<td></td>
```
davecgh: 2 data pushes.  1st push is DER signature.  2nd push is public key.

docs/advanced/transaction-details.md
```diff
@@ -75,3 +76,210 @@ The witness data of a transaction involves only its inputs. The included data fi
...
+</tr>
+<tr>
+<td>Signature script length</td>
+<td><code>6b</code></td>
+<td>107</td>
+</tr>
+<tr>
+<td>Signature script</td>
+<td style="word-break: break-all;"><code>483045022100da5b0fde58c4c57a88d96f83b02eddcb67b1b68ce95cc562c895f2e4e57bb44302201a8744aed654d2979bdd797187f6bf63ad1898c5d5c585309eeca558742858b201210307e3d98b004b15561c0f5aa158c90f93de33756d3bfd1f8a1058a4d90ffc7ec7</code></td>
+<td></td>
```
davecgh: 2 data pushes.  1st push is DER signature.  2nd push is public key.

docs/advanced/transaction-details.md
```diff
@@ -75,3 +76,210 @@ The witness data of a transaction involves only its inputs. The included data fi
...
+</tr>
+<tr>
+<td>Tree</td>
+<td><code>00</code></td>
+<td>0</td>
+</tr>
+<tr>
+<td>Sequence number</td>
+<td><code>ffffffff</code></td>
+<td>4294967295</td>
```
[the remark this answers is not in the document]
s-ben: Added `(transaction finalized)`

docs/advanced/transaction-details.md
```diff
@@ -75,3 +76,210 @@ The witness data of a transaction involves only its inputs. The included data fi
...
+</tr>
+<tr>
+<td>Output index</td>
+<td><code>01000000</code></td>
+<td>1</td>
+</tr>
+<tr>
+<td>Tree</td>
+<td><code>00</code></td>
+<td>0</td>
```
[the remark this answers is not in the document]
s-ben: Added `(use regular transaction fee)`

docs/advanced/transaction-details.md
```diff
@@ -75,3 +76,210 @@ The witness data of a transaction involves only its inputs. The included data fi
...
+<tr>
+<td colspan="3" style="text-align: center;">1st output</td>
+</tr>
+<tr>
+<td>Output amount</td>
+<td><code>91c2e26d02000000</code></td>
+<td>104.33</td>
+</tr>
+<tr>
+<td>Version</td>
```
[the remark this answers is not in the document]
s-ben: Added

docs/advanced/transaction-details.md
```diff
@@ -75,3 +76,210 @@ The witness data of a transaction involves only its inputs. The included data fi
...
+</tr>
+<tr>
+<td>Public key script length</td>
+<td><code>19</code></td>
+<td>25</td>
+</tr>
+<tr>
+<td>Public key script</td>
+<td style="word-break: break-all;"><code>76a914f2ccddc70f8b8bcea24a362bf404841d53cbebc088ac</code></td>
+<td></td>
```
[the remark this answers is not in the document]
s-ben: Added. Used this format,
`OP_DUP OP_HASH160 f2ccddc70f8b8bcea24a362bf404841d53cbebc0 OP_EQUALVERIFY OP_CHECKSIG` 
`(Pay-To-Pubkey (P2Pk) hash script)`

docs/advanced/transaction-details.md
```diff
@@ -75,3 +76,210 @@ The witness data of a transaction involves only its inputs. The included data fi
...
+<tr>
+<td>Public key script length</td>
+<td><code>19</code></td>
+<td>25</td>
+<td></td>
+</tr>
+<tr>
+<td>Public key script</td>
+<td style="word-break: break-all;"><code>76a9141a1f5c7e9f7696989dced5cb9b61464f95790b7288ac</code></td>
+<td></td>
```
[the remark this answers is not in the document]
s-ben: Added. Same format as above comment.

docs/advanced/transaction-details.md
s-ben: Updated all numbers to 8 decimal places.

docs/advanced/transaction-details.md
s-ben: Updated

docs/advanced/transaction-details.md
s-ben: Updated

docs/advanced/transaction-details.md
s-ben: Just used this text as is. This is definitely helpful, though wondering, would it be useful to break this down further? E.g. separate out DER signature and public key hex values and label them? Or can we assume the reader knows how to interpret this?

docs/advanced/transaction-details.md
s-ben: Same comment as above

s-ben:
@davecgh have implemented all your suggested changes, as well as tweaked the formatting to accommodate some of the new long strings (working on FF as well). My only outstanding question is on the Signature Script description (in comment on suggestion).

docs/advanced/transaction-details.md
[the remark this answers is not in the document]
davecgh: tree, not fee

docs/advanced/transaction-details.md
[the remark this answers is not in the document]
davecgh: That's fine, but it needs to be `(Pay-To-Pubkey-Hash (P2PKH) script)`.  It's actually paying to the pubkey hash, not the pubkey.

docs/advanced/transaction-details.md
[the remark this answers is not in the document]
davecgh: Same here regarding paying to the pubkey hash.

docs/advanced/transaction-details.md
davecgh: If it will fit well, that makes sense.  I was trying to keep it relatively short because the format I see on the current markdown preview is wonky.

If you want to break it down more though, it's:

2 data pushes.

| Script Operation | Value | Description|
| --- | --- | --- |
| 1st push |`3045022100da5b0fde58c4c57a88d96f83b02eddcb67b1b68ce95cc562c895f2e4e` `57bb44302201a8744aed654d2979bdd797187f6bf63ad1898c5d5c585309eeca558` `742858b201` | DER signature |
| 2nd push | `0307e3d98b004b15561c0f5aa158c90f93de33756d3bfd1f8a1058a4d90ffc7ec7` | Public key |

davecgh:
I still see several unaddressed review items.  They are collapsed.

docs/advanced/transaction-details.md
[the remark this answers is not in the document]
s-ben: Updated

docs/advanced/transaction-details.md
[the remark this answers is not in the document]
s-ben: Fixed

docs/advanced/transaction-details.md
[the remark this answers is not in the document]
s-ben: Fixed

docs/advanced/transaction-details.md
s-ben: Formatting is getting a little wonky, but my guess is that if you're slogging through this level of detail, you'll probably appreciate more breakdown. 

Below is a screenshot of how I've implemented it (FF),
image

Let me know if this text is inaccurate, or could be worded more clearly. I'm still slightly confused as to what the DER signature is. I assume it's a binary string of digits encoded with DER, which is what `asm` is in the decoded JSON (i.e. this is still encoded),

```
         "scriptSig": {
            "asm": "3045022100da5b0fde58c4c57a88d96f83b02eddcb67b1b68ce95cc562c895f2e4e57bb44302201a8744aed654d2979bdd797187f6bf63ad1898c5d5c585309eeca558742858b201 0307e3d98b004b15561c0f5aa158c90f93de33756d3bfd1f8a1058a4d90ffc7ec7",
            "hex": "483045022100da5b0fde58c4c57a88d96f83b02eddcb67b1b68ce95cc562c895f2e4e57bb44302201a8744aed654d2979bdd797187f6bf63ad1898c5d5c585309eeca558742858b201210307e3d98b004b15561c0f5aa158c90f93de33756d3bfd1f8a1058a4d90ffc7ec7"
         }
```

docs/advanced/transaction-details.md
s-ben: Same comment as above. Note that the data and from above was from the 2nd signature script, even though comment was on the 1st. 

Tried to follow what you did when breaking down the other signature. Should definitely be reviewed for accuracy though.

s-ben:
> I still see several unaddressed review items. They are collapsed.

Are there wrongly resolved comments? Have addressed the outstanding comments I can see.

docs/advanced/transaction-details.md
davecgh: That text is fine except it's just a DER signature.  No need to say asm.  asm just means it's the disassembly of the script.

There really is no need to break the DER signature down even more in this document imo.  It's a well known and defined ASN.1 encoding.

docs/advanced/transaction-details.md
davecgh: Reviewed.  It's accurate.

davecgh:
Reviewed updates, resolved relevant comments, and answered questions.

If you look, there is a "7 hidden conversations" in the middle where the review comments are hidden until you click the "Load more...".  Those are the unaddressed items I was referring to.

docs/advanced/transaction-details.md
```diff
@@ -75,3 +76,210 @@ The witness data of a transaction involves only its inputs. The included data fi
...
+<tr>
+<td>Version</td>
+<td><code>0000</code></td>
+<td>0</td>
+</tr>
+<tr>
+<td>Public key script length</td>
+<td><code>19</code></td>
+<td>25</td>
+<td></td>
```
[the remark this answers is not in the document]
s-ben: Removed

docs/advanced/transaction-details.md
s-ben: Noted:
```
0
(no expiry)
```

docs/advanced/transaction-details.md
s-ben: Noted:
```
0
(no lock time)
```

docs/advanced/transaction-details.md
s-ben: Removed

docs/advanced/transaction-details.md
s-ben: Removed

docs/advanced/transaction-details.md
s-ben: Yes, I think we have drilled far enough. Removed `(asm)`.

s-ben:
Found the "7 hidden conversations" in the middle, but when I click on them it just expands to show already resolved changes :/ Do you see any unresolved changes still on your end? I hate this UI feature.

davecgh:
I resolved them because you addressed them in the latest commit.  I approved the PR already.

s-ben:
Cool. Thx for the thorough review/patience on this one 👍

PR #951, opened 2019-05-28
Flesh out transaction details page

### LINKED ISSUE #942: Document Decred-specific transactions
On the [Transaction Details](https://docs.decred.org/advanced/transaction-details/) page, we provide a general transaction specification. This is sufficient for documenting common transactions (`Pay-to-Pubkey`, `Pay-to-PubkeyHash`, etc.), as the lower-level format details should be the same as for Bitcoin. However, Decred-specific transactions, such as Stakebase transactions related to tickets, are not currently documented anywhere but comments in dcrd. A developer not familiar with Decred may find useful more documentation on how these Decred-specific transactions work. E.g. an exact byte-map for a transaction that buys a ticket.

s-ben:
Inspired by this [chat](https://matrix.to/#/!tfqymymiNgzSUJTHqS:decred.org/$1558655816706MGlkN:decred.org?via=decred.org&via=matrix.org&via=zettaport.com) in #Documentation with @davecgh and @jrick, this PR seeks to flesh out our documentation of Decred transactions.

The following additions have been made to the [Transaction Details](https://docs.decred.org/advanced/transaction-details/) page:

* Adds section defining Decred-specific transaction types ( Closes #942 ). Note: I'm reusing the definitions we came up with for the glossary -- these definitions always seemed too technical for a high-level glossary anyway, so I think the full definitions belong here. We could shorten the definitions in the glossary and point to this page for further details if we want to avoid duplicate content. 
* Adds text further explaining what the transaction format/spec does (and does not) define, as well as instructions on how to find lower-level documentation of signature types in dcrd.
* Adds 'Bitcoin similarities' section, in case the reader is familiar with Bitcoin (or relying on Bitcoin docs as a starting point). Not terribly attached to this one, can delete if we don't think it's useful.


docs/advanced/transaction-details.md
```diff
@@ -2,11 +2,28 @@
...
+* **Stakebase transaction:** Every block which is mined will contain a stakebase transaction for each ticket which voted on that block. Stakebase transactions are located in the stake transaction tree and have two inputs: the Decred which was spent to purchase the ticket, and the newly created Decred constituting the reward for voting. Stakebase transactions are also created when an expired or missed ticket is revoked, however there will be no voting reward created in these cases. Stakebase transactions belong to the stake transaction tree, which means that these transactions cannot be rejected by PoS voters, even if they vote to reject the block which contains them.
```
davecgh: Every block starting at the stake validation height which is mined ...

docs/advanced/transaction-details.md
```diff
@@ -2,11 +2,28 @@
...
+
+
+## Transaction Types
+
+To enable Decred's consensus rules, transactions are split into two separate trees within a block: the regular transaction tree and the stake transaction tree. The stake transaction tree contains transactions related to ticket buying and ticket voting rewards. The regular transaction tree contains all other transactions. 
+
+In addition to regular transactions (sending DCR from one address to another), there are a few special transaction types to be aware of:
+
+* **Coinbase transaction:** Every block which is mined contains a single coinbase transaction. This transaction will only have one input, and that is newly created Decred which constitutes the PoW and Treasury portions of the block reward. Coinbase transactions are part of the regular transaction tree, which means that they will be rejected if Proof-of-Stake (PoS) voters vote to reject the block which contains them. Decred created in coinbase transactions cannot be spent until the coinbase maturity period has passed.
```
davecgh: Inputs can't create coins.

docs/advanced/transaction-details.md
davecgh: Same here.  The inputs don't create the coins, rather it's the outputs.

davecgh: s/reject/disapprove/

docs/advanced/transaction-details.md
```diff
@@ -2,11 +2,28 @@
...
+* **Multisignature:** Multisignature refers to transactions which can be authorized by more than one private key. Multisignature transactions can support multiple keys (N) and a subset of those (M) are required to transact (commonly known as “MofN”). For example, a 2 of 3 multisignature transaction would have three valid keys, however only two of them would be required to authorize. Multisignature transactions are in the regular transaction tree.
 
 
 ## Transaction Format 
 
+At the protocol level, all transactions share the same basic transaction format; the transaction type is determined by the signature type present in the transaction. This section below describes this basic transaction format.
```
davecgh: This might be acceptable, but "The section" sounds better to me.

s-ben: Good catch. Changed to "The section"

docs/advanced/transaction-details.md
s-ben: Changed to 'output',

> Every block which is mined contains a single coinbase transaction. This transaction will only have one **output**, and that is...

docs/advanced/transaction-details.md
s-ben: The GitHub UI is lumping all three comments together, making it difficult to see which line each comment is pointing to, so just going to paste the edits I made below, in case anything jumps out as still wrong. Bold is to highlight changes, not in actual page. 

>Stakebase transaction: Every block **starting at the stake validation height** which is mined

>This transaction will only have one **output**, and that is newly created Decred which constitutes the PoW.. 

This one is made in two places:

>Coinbase transactions are part of the regular transaction tree, which means that they will be rejected if Proof-of-Stake (PoS) voters vote to **reject/disapprove** the block which contains them.

>Stakebase transactions belong to the stake transaction tree, which means that these transactions cannot be rejected by PoS voters, even if they vote to **reject/disapprove** the block which contains them.

docs/advanced/transaction-details.md
davecgh: Unfortunately, that is not true either.  There are many consensus rules around coinbase transactions, but in terms of the number of inputs and outputs, a coinbase transaction is technically only required to have 1 null input, which indicates no existing coins are being spent, and 2 outputs.

The first output creates the coins that are added to the project treasury and is required to be 10% of the total generated block subsidy for that height.  The second output contains the encoded block height which ensures coinbase hash collisions are impossible.

After that, there can be an arbitrary number of additional outputs (up to the maximum limits) so long as the sum of the remaining outputs does not pay more coins than 60% of the total generated block subsidy for that height.

Consequently, a typical coinbase consists of 3 outputs, the required aforementioned 2 outputs and a the third output pays the 60% PoW subsidy portion to the PoW miner.

docs/advanced/transaction-details.md
davecgh: That was programmer speak for replace "reject" with "disapprove".  You can't vote to reject a block, only to disapprove it.  Actually rejecting a block is accomplished by not casting a vote at all.

docs/advanced/transaction-details.md
s-ben: While this is all good info, wary of getting too granular here. Someone really needing this info will probably end up in the source code anyway (e.g. noting that non-typical transactions might have multiple outputs instead of just the output sending 60% to one miner). Also, changing block subsidy to block reward for consistency with other high-level docs. How does this sound?

>Coinbase transaction: Every block which is mined contains a single coinbase transaction. A coinbase transaction will typically have three outputs: one output that contains the encoded block height (used to ensure that coinbase hash collisions are impossible), one output that creates the coins added to the project treasury (10% of the block reward), and one output that creates the coins paid to the PoW miner for creating the block (60% of the block reward). Coinbase transactions are part of the regular transaction tree, which means that they will be rejected if Proof-of-Stake (PoS) voters vote to disapprove the block which contains them. Decred created in coinbase transactions cannot be spent until the coinbase maturity period has passed.

docs/advanced/transaction-details.md
s-ben: Ah. Added to my developer to English dictionary. 

>Actually rejecting a block is accomplished by not casting a vote at all.

Good to know. There was some confusion about this terminology on Matrix the other day.

docs/advanced/transaction-details.md
davecgh: Sounds good and is accurate.  Nice job distilling it down to the most relevant bits!

docs/advanced/transaction-details.md
```diff
@@ -2,11 +2,28 @@
...
+* **Stakebase transaction:** Every block starting at the stake validation height which is mined will contain a stakebase transaction for each ticket which voted on that block. Stakebase transactions are located in the stake transaction tree and have two outputs: the Decred which was spent to purchase the ticket, and the newly created Decred constituting the reward for voting. Stakebase transactions are also created when an expired or missed ticket is revoked, however there will be no voting reward created in these cases. Stakebase transactions belong to the stake transaction tree, which means that these transactions cannot be rejected by PoS voters, even if they vote to disapprove the block which contains them.
```
davecgh: It's not true that stakebase transactions only contain two outputs.  It's really quite complicated and there are a lot of rules around it.  I'm not sure it's really appropriate to call that out here as that seems more suitable for developer docs.

I would say perhaps just:

>Every block starting at the stake validation height which is mined will contain a stakebase transaction for each ticket which voted on that block. Stakebase transactions are located in the stake transaction tree and consist of a very specific form which identifies the ticket that is associated with it as well as produces the Decred which was spent to purchase said ticket and the newly created Decred constituting the reward for voting.  Stakebase transactions are also created when an expired or missed ticket is revoked, however there will be no voting reward created in these cases. Stakebase transactions belong to the stake transaction tree, which means that these transactions cannot be rejected by PoS voters, even if they vote to disapprove the block which contains them.

docs/advanced/transaction-details.md
s-ben: Thx! Your writing and feedback is definitely helping me understand the concepts

docs/advanced/transaction-details.md
davecgh: Also, I just noticed there is a repetitive bit in there "Stakebase transactions are located in the stake transaction tree ...." and later "Stakebase transactions belong to the stake transaction tree, which means ...."

s-ben: Reworded for flow:

>Stakebase transaction: Every block starting at the stake validation height which is mined will contain a stakebase transaction for each ticket which voted on that block. The stakebase transaction will identify the ticket that voted, as well as produce the Decred which was spent to purchase the ticket, and the newly created Decred constituting the voting reward. Stakebase transactions belong to the stake transaction tree, which means that these transactions cannot be rejected by PoS voters, even if they vote to disapprove the block which contains them.

This does raise a question I've had for a while though: when you say the stakebase transaction "produces" the Decred which was spent to purchase the ticket, do you mean that it "unlocks" Decred that was locked in a smart contract when the ticket was purchased? Or does it create Decred that was destroyed? I assume the former.

davecgh: I mean that it actually produces new coins.  Under the hood, the act of acquiring a ticket destroys the Decred until either a vote or a revocation mints the same amount in new coins to replace them.  The documentation refers to this process as "locking" because it's conceptually easier to understand and the coins are indeed not spendable until a vote or revocation effectively "unlocks" them.

davecgh: As far as the proposed text, it sounds good and is accurate.

s-ben: Interesting. Ok I think we're good to merge then unless anyone else chimes in.

PR #952, opened 2019-05-29
bump mkdocs-material to 4.3.1

dajohi:


s-ben:
Built locally and all the pages look fine (assuming that is all that needs doing here).

PR #953, opened 2019-05-29
1st draft LN page

s-ben:
This PR creates a simple 'Lightning Network (LN)' page under wallets. There's a lot of info I can think to add here (Decredition support, how to set up and operate nodes, etc.), but I realized we probably don't want to add that to dcrdocs until there's an official release of Decrediton that supports LN. We could start adding more technical details in a new page under '/Advanced', but a page under '/Wallets' should be more end user oriented. 

I'm tempted to say let's hold off on this page until there's something a user can use without having to compile from source, but it's also awkward that we're promoting our Lightning launch and don't have it in the docs at all. Thoughts?

### DIFF
--- a/docs/wallets/lightning.md
+++ b/docs/wallets/lightning.md
@@ -4,7 +4,7 @@
 
 ## What is Lightning (LN)?
 
-The Lighning Network (LN) is a "Layer 2" [protocol](https://en.wikipedia.org/wiki/Lightning_Network) that runs on top of the base layer Decred blockchain ("Layer 1"). It enables fast, low cost transactions and higher throughput than possible on Layer 1. 
+The Lightning Network (LN) is a "Layer 2" [protocol](https://en.wikipedia.org/wiki/Lightning_Network) that runs on top of the base layer Decred blockchain ("Layer 1"). It enables fast, low cost transactions and higher throughput than possible on Layer 1. 
 
 ## Decred Lightning Implemetation
 
--- a/docs/wallets/lightning.md
+++ b/docs/wallets/lightning.md
@@ -4,7 +4,7 @@
 
 ## What is Lightning (LN)?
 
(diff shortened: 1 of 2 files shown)


RichardRed0x:
I like the idea of a more detailed page under Advanced which is about getting people up to speed with how they can get on testnet, and what they can do on it. At a point where there's significant mainnet LN action, some of the content could be recycled for the wallets page.

s-ben:
Agreed. For instance, I know there's some advanced instructions on using LN wallet already (e.g. not reusing the same seed from your regular Decredition wallet), which could be migrated to /Wallet. However it seems most of this stuff is in flux, so hesitant to put too much effort in just yet.

s-ben:
I think we should at least get up a simple landing page though. If people are coming to the docs just to check out the project, they should at least be aware we have LN. If people come to the docs because they saw that we had LN, we should at least have a page pointing them to the dcrlnd repo for more info.

docs/wallets/lightning.md
```diff
@@ -0,0 +1,13 @@
+# Lightning Network (LN)
+
+---
+
+## What is Lightning (LN)?
+
+The Lightning Network (LN) is a "Layer 2" [protocol](https://en.wikipedia.org/wiki/Lightning_Network) that runs on top of the base layer Decred blockchain ("Layer 1"). It enables fast, low cost transactions and higher throughput than possible on Layer 1.
```
fguisso: LN is a Layer 2 Protocol, if the paragraph refers generaly LN, you need to say: `that runs on top of blockchain-based cryptocurrency (like Decred)`

s-ben: Good suggestion. Went with:

>The Lightning Network (LN) is a "Layer 2" [protocol](https://en.wikipedia.org/wiki/Lightning_Network) that runs on top of the base layer ("Layer 1") of a blockchain-based cryptocurrency (like Decred).

PR #958, opened 2019-06-22
Rename DAE to DAO

### LINKED ISSUE #957: Change DAE (Decred Autonomous Entity) to DAO (Decred Autonomous Organization)
As noted in this [Issue](https://github.com/decredcommunity/issues/issues/134), there have been several conversations about renaming the Decred DAE (Decentralized Autonomous Entity) to the Decred DAO (Decentralized Autonomous Organization), with the general consensus being that we should rename to DAO. This concensus was further expressed in a [reddit post](https://www.reddit.com/r/decred/comments/c02geh/poll_rename_dae_to_dao/) and recent [chat](https://matrix.to/#/!tIDEIWechmqCLjPiui:decred.org/$1560971319936QXqBL:decred.org?via=decred.org&via=matrix.org&via=zettaport.com) in #Governance. 

When we originally went with DAE, it was to avoid association with the ETH DAO hack. Since then, DAOs have becoming a hot, trending topic in crypto, and not associated with the DAO hack. In order to be part of the broader conversation around DAOs (and avoid extra overhead in explaining a term Decred invented), I propose changing DAE to DAO here and in all Decred properties. 

Creating this issue to see if there is anyone that opposed this change that did not have a chance to participate in the above referenced chats on reddit and Matrix.

s-ben:
This PR renames the Decred 'Decentralized Autonomous Entity (DAE)' to the Decred 'Decentralized Autonomous Organization (DAO)' ( Closes #957 ).


s-ben:
>The first is an interesting case because it appears in the constitution which has only been modified in the past after having been approved by stakeholders. I believe changing this without a proposal is acceptable as this kind of change is not substantial - it is akin to fixing a typo.

We discussed putting this name change up for a vote on Politeia. It was decided that unless there was significant opposition to renaming, it would be a waste of time doing a Pi proposal. So far nobody on chat, reddit, or GitHub has voiced opposition. I think a non-controversial name change in the constitution is defensible (and reversible if someone really feels strongly about it). Fine with not changing it there as well. @jholdstock

PR #961, opened 2019-06-28
Fix bug restore wallet Decrediton instructions

### LINKED ISSUE #959: Restore Decrediton VSP prompt
When restoring an existing wallet in Decrediton 1.4.0, no prompt appears requesting a VSP key. Docs mention a prompt. 

https://docs.decred.org/wallets/decrediton/decrediton-setup/#restore-existing-wallet

Proper process is to manually import the redeem script, which can be found by logging in to your VSP account.

s-ben:
Fixes a bug in the steps for restoring a wallet ( Closes #959 ). Before, we said the user would be prompted to enter their VSP's redeem script. Actually there is not prompt and they need to manually import it after restoring their wallet.

Have also added a step to close an existing wallet (if open) to get to the `Restore Existing Wallet` prompt. It was kind of difficult to find the `Close Wallet` button in the UI, so have added a screenshot showing where it is. Can remove if we think text is enough.


jholdstock:
Apologies, I was actually incorrect when writing up the issue #959.

Proper process is actually to find the API key in the VSP, and then copy that into decrediton. Its actually the same process to set up a new wallet or to restore an existing wallet. Redeem script is not the recommended process, that is just for emergencies (eg. API key is not retrievable, VSP is offline)

s-ben:
No worries. Easy change. Here's what we have in the commit I just pushed,

> * If you need to reconnect to your Voting Service Provider (VSP), you'll need to re-import your VSP API key, as described in the Tickets section on the [Using Decrediton](using-decrediton.md) page.

PR #963, opened 2019-06-29
Adds page pseudorandom number generation algorithm

### LINKED ISSUE #941: Add random number generator algorithm details
In a recent Reddit [post](https://www.reddit.com/r/decred/comments/b8k2i0/where_can_i_read_about_the_pseudo_random_number/), @davecgh answered a question about the pseudo random number generator algorithm used to selects tickets. Dave's description of the algorithm is pretty solid and could probably be dropped into the docs basically as is. I think that if one person had this question, others might as well, and that this information would be good to include somewhere. Thoughts? Does anyone know if this is already documented elsewhere (e.g. a research paper we're already linking to)? Suggestions for a good place to put this?

s-ben:
This PR adds a page under /research (open to placing it elsewhere) describing the details of the algorithm used to pseudorandomly select tickets for voting ( Closes #941 ). 

@davecgh, I'm pretty much just rewording your original [reddit post](https://www.reddit.com/r/decred/comments/b8k2i0/where_can_i_read_about_the_pseudo_random_number/) for flow. Do let me know if I've somehow mangled it. Also, if there is any other relevant info you can think of (e.g. reasoning for why this algo was chosen, references, etc.), can add.


docs/research/ticket-selection-algorithm.md
```diff
@@ -0,0 +1,8 @@
...
+
+## Pseudorandom number generation
+
+Tickets are selected from the ticket pool using a deterministic [Pseudorandom Number Generator (PRNG)](https://en.wikipedia.org/wiki/Pseudorandom_number_generator) algorithm. It is deterministic because the seed input to the PRNG algorithm is derived from the hash of the previous block (the block being voted on). Specifically, the seed is a serialization of the [header](../advanced/block-header-specifications.md) of the previous block suffixed with a constant derived from the hex representation of Pi, which acts as a publicly verifyable Nothing-Up-My-Sleeve ([NUMS](https://en.wikipedia.org/wiki/Nothing-up-my-sleeve_number)) constant. All eligible tickets in the ticket pool (also referred to as live tickets) are hashed using the [BLAKE-256](blake-256-hash-function.md) hash function and sorted [lexicographically](https://en.wikipedia.org/wiki/Lexicographical_order) by their hash to generate a [total order](https://en.wikipedia.org/wiki/Total_order), with the total number of eligible tickets as an upper bound. Finally, uniformly random values are produced and used as indicices into the total order.
```
davecgh: The word algorithm shouldn't be after PRNG in the first sentence.  "using a deterministic Pseudorandom Number Generator algorithm" doesn't read properly.  Instead just "using a deterministic Pseudorandom Number Generator" is accurate.

davecgh: Same goes later for PRNG algorithm.  Just PRNG.

davecgh: verifiable

davecgh: Perhaps clarifying Pi here as

> the mathematical constant Pi (𝜋)

would be helpful.

davecgh: > by their hash to generate a total order, with the total number of eligible tickets as an upper bound.

That is not correct since the upper bound doesn't apply to the total order itself rather to the uniformly random values.  How about:

>Finally, uniformly random values constrained by the number of total eligible tickets as an upper bound are produced and used as indices into the total order.

Also note that "indicices" is not spelled correctly in the original text.

s-ben:
@davecgh, comments addressed. Suggested text lgtm.

docs/research/ticket-selection-algorithm.md
```diff
@@ -0,0 +1,8 @@
+# <img class="dcr-icon" src="/img/dcr-icons/QuestionTicket.svg" /> Ticket Selection Algorithm
+
+Every block, 5 tickets are pseudorandomly selected from the ticket pool to [vote](../../proof-of-stake/overview/). Below is a high-level description of the ticket selection algorithm. A more detailed, low-level walkthrough of the ticket selection code can be found in this [video](https://www.youtube.com/watch?v=eysGWVhDFWY).
```
jholdstock: link should be to a `.md` file

s-ben: Fixed. FYI, I do remember trying `.md` before and it wasn't working, so went with the `/`. I think my mistake was that with the `/`, I was able to use a relative path that wasn't technically correct (`../../proof-of-stake/overview/`) and it worked. Using `.md` with that path doesn't work. If you use the correct relative path (`../proof-of-stake/overview.md`) it does.

jholdstock: Exactly its a relative path from the current source file to target source file. This looks good now.

PR #967, opened 2019-07-13
Add atomic swaps to docs

### LINKED ISSUE #964: Add atomic swaps to docs
Decred has done considerable work on atomic swaps, and the [decred/atomicswap](https://github.com/decred/atomicswap) repo is under active development. There also exists a considerable amount of technical documentation in the repo README. While I think it makes sense to leave most of the technical documentation on GitHub (where most devs will find it, it's one of our most starred/forked repos), it seems atomic swaps and how they work with Decred specifically, should be represented somewhere in the docs. I would propose a page under /advanced or /research, with the following elements:

- High-level summary of what an atomic swap is
- Basic description of what you can do with Decred's atomic swap tools
- (optional) history of Decred's contribution to atomic swaps
- Pointer to the decred/atomicswaps repo for further technical documentation

imestin:
Closes #964

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,6 +42,7 @@ INFO    -  Cleaning site directory
 [I 160402 15:50:43 handlers:58] Start watching changes
 [I 160402 15:50:43 handlers:60] Start detecting changes
 ```
+
 If you are using Windows, you may need to inform python to search sys.path for the mkdocs module:
 
 ```bash
@@ -52,6 +53,7 @@ INFO    -  Cleaning site directory
 [I 190207 18:05:35 handlers:59] Start watching changes
 [I 190207 18:05:35 handlers:61] Start detecting changes
 ```
+
 Open up <http://127.0.0.1:8000> in your browser, and you will see the documentation home page being displayed. The dev-server also supports auto-reloading, and will rebuild your documentation whenever anything in the configuration file or the documentation directory changes.
 
 ## Deploying
@@ -94,6 +96,7 @@ $ docker run -d --rm -p <local port>:80 decred/dcrdocs:latest
 
 - `wallets/decrediton/decrediton-setup.md`
 - `wallets/decrediton/using-decrediton.md`
+- `wallets/decrediton/decrediton-troubleshooting.md`
 
 #### gominer release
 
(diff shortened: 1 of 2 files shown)


s-ben:
@imestin, first pass it looking good. A few suggested changes. 

* This PR is from your local master branch. Typically when you make changes, you want to create a local feature branch from master first (e.g. `add-atomic-swap`), then submit a the PR from that feature branch. This is common practice and makes it easier to keep track of branches. Also, while I was able to pull this and get it running locally, I think it wrote over my local master branch, causing a little mess. Not hard to fix, but just something to keep in mind.
* Your changes are spread across multiple commits. This is usually OK when iterating on changes with reviewers, or when there are only a couple commits. But generally, if you create a number of commits while working on something (common), you'll want to `git squash` your commits down to a single commit before submitting the PR. This makes it easier to review, as once this is in a single commit, I can use the Review feature to comment directly on lines. @jholdstock is this correct policy?
* The `atomic-swap` page needs to be added to the navigation by updating `mkdocs.yaml`

>Later, Atomic Swaps will be integrated in the Decredition GUI wallet.

It’s generally a bad idea to say things will be integrated in the future, even when we’re almost certain they will. Docs should only document what already exists. Otherwise, either a) it doesn’t get integrated somehow and the docs are wrong, or (more likely), b) it does get integrated but we forget to update this page, making people think it’s not integrated yet. 

>The first atomic swap happened between Decred and Litecoin on September 20, 2017.

Is there an interesting link we can insert here?

> For a decentralized exchange to work, Lightning Network is also necessary. 

Can you say why this is true? Not absolutely necessary, just curious, and so might some other readers.

jholdstock:
Hi @imestin. Can you swap these over please?

Screenshot from 2019-07-14 09-45-38

jholdstock:
Its best to use common sense when structuring your PR commits.  Sometimes squashing down to fewer commits can be helpful, sometimes it can hide important details.

If you are able to break the PR down into multiple commits which make sense on their own, and can be reviewed in isolation, thats good. Smaller chunks are good. 

If you have a bunch of commits like "fix this" "fix that" "oops, revert previous", "try again" -  this is just noise and does not help anybody.

imestin:
Thank you for feedback, tomorrow I will correct the errors.

imestin:
> Can you say why this is true? Not absolutely necessary, just curious, and so might some other readers.

Yes, you are right, that was a bad assumption from me, other tools can do this as well.

imestin:
I created another pull request that only has 1 commit and fixes the above errors.

Issue #816, opened 2019-01-27
Dedicated Privacy page


xaur:
Collect all privacy gotchas (theory) and best practices on one page.

Some ideas for content:

* tradeoffs of reusing vs not reusing addresses: simplicity and transparency (sometimes needed) vs complexity and privacy, payment protocols
* spending reveals pubkey, key/address that never spent is better protected from QC (perhaps this belongs to security, but privacy is very related too)
* not leaving identifiers in public places
* privacy implications of checking one's data via public block explorer
* privacy of dcrwallet/decrediton's SPV and of 3rd party wallets
* privacy degradation from using centralized KYCed third parties
* privacy implications of using atomicswap tools

xaur:
* privacy of using Politeia, trickling, Tor
* using email address on VSPs and Politeia

xaur:
* privacy loss from reusing voting address for different tickets when using VSP

imestin:
i would like to work on this.

jholdstock:
Hi @imestin, fine for you to work on this, but I would suggest working on only a few of these topics at a time. Many small PRs are preferred to one large PR - it makes reviewing much easier, reduces chance of mistakes being made and conflicts appearing.

imestin:
ok, I will do it in small pieces.

xaur:
Some huge resources to mine for privacy knowledge:

* https://en.bitcoin.it/wiki/Privacy
* https://github.com/6102bitcoin/FAQ/blob/master/hodl-privacy.md
* https://medium.com/@nopara73/coin-control-is-must-learn-if-you-care-about-your-privacy-in-bitcoin-33b9a5f224a2

Not everything is applicable to Decred until our wallets gain more coin control features, but I suspect a lot of general and common sense advice can be reused or at least referenced.

imestin:
How deep this docs should be? (compared to https://en.bitcoin.it/wiki/Privacy for example) should I redirect to outside resources and keep the length of the doc below 1/2 pages?

xaur:
Great question @imestin

I think it can start as an overview and list of high level directions written in simple language.

- page that turns user's privacy brains on and brings all the nuance into his awareness
- put lowest hanging fruit first to help us mortals start at least somewhere
- go deeper into things specific to Decred
- overview and link to all Decred's privacy features
- don't hesitate to link to great outsude resources
- consider expanding this issue from one page to a group of pages

[...]

PR #986, opened 2019-09-10
Add Privacy to docs

imestin:
Closes Dedicated Privacy page #816

### DIFF
--- a/docs/privacy/general-privacy.md
+++ b/docs/privacy/general-privacy.md
@@ -1,0 +1,60 @@
+# General Information on Privacy
+
+## About Transaction Graphs
+
+It’s important to understand, that transactions are pseudonymous, but not anonymous. That means, a transaction graph can be built up, and there exists companies that are doing that work. If at some point one of your transactions is connected to your real world identity, it becomes easy to make assumptions based on that, for example that other transactions also belong to you, or that a specific unspent transaction belongs to you (that could give information about your balance). 
+One possible way your personal identity could be linked to your blockchain activity is through the use of a block explorer. The administrator of the block explorer may be able to view your IP address and which addresses or transactions you are browsing, and they may infer that you own those addresses or transactions. A possible solution for this is to use blockchain explorers exclusively over Tor.
+
+## Transaction Broadcasting
+
+Broadcasting a transaction can reveal that the transaction you are broadcasting belongs to you. A solution for this is to broadcast your transaction through a website, over Tor. For example, <https://explorer.dcrdata.org/decodetx>
+It's also important to note, that different wallets might use different [address types](../advanced/address-details.md). This can also give a hint about your identity, which combined with other signals can ultimately reveal your identity.
+
+## Trade-offs of Reusing vs Not Reusing Addresses
+
+For privacy the best practice is to not reuse an address ever. A cryptocurrency address can be thought of as a cryptocurrency invoice instead. The term "address" does not mean, that it was designed to reuse it, like one would do with a bank account number for example. A privacy conscious user will use a newly generated address for every single transaction, and never reuse an address unless absolutely necessary. 
+On the other hand, there may be situations where address reuse may be preferred, or even necessary. The Decred Treasury is a good example of this. Transparency and auditability are more important than privacy in the context of the treasury.
+
+## Change Address
+
+If we only spend a fraction of an Unspent Transaction Output ([UTXO](https://coincentral.com/utxo-beginners-explainer/)), a change output will be generated. This is not good for privacy. For example, the change UTXO will be merged with another unspent transaction output next time when we are spending money, thus creating a signal that both of the transactions belong to us. An average wallet software won’t give us information about this, we won’t know, for example, that 2 UTXOs were merged to create the transaction. If you care about privacy, you should exactly know which UTXOs are being spent. [This](https://medium.com/@nopara73/coin-control-is-must-learn-if-you-care-about-your-privacy-in-bitcoin-33b9a5f224a2) is a good article on coin control.
+
+## Avoid Identification when Acquiring Cryptocurrencies 
+
+Try to avoid getting cryptocurrencies from websites that require KYC or AML, or any other means of identification. Be aware that centralized exchanges can be hacked, thus providing the collected information to the hacker, who can belong to any kind of organization. It’s also possible that at some point an oppressive government will request the collected information. ATMs can be equipped with cameras, or there can be cameras nearby the ATM. This combined with the blockchain can reveal your identity. 
+
+## Avoid Leaving Identifiers in Public Places
+
+Many times this is something physical, for example leaving behind a printed QR code somewhere, writing your nickname or real name on a piece of paper. Your name can be inprinted on the next page of the sticky notes stack, for example. Or that could also happen, that you leave behind your phone or computer unlocked for some minutes, with important information displayed on screen.
+
+## Centralized Mixers
+
+The aim of a cryptocurrency mixer is to break the transaction graph. An old version of cryptocurrency mixers is the centralized mixers (also called scrambler, thumbler, shuffler). 
+Either we use a centralized mixer or the later mentioned solutions, [amount correlation](https://en.bitcoin.it/wiki/Privacy#Amount) is a problem.
+
+## Non-custodial Coin Mixing
+
+It is possible to mix coins without a centralized exchange. One method is [TumbleBit](https://github.com/decred/tumblebit), which allows parties to make payments through an untrusted Tumbler.
+[Decred's build in coin mixing](https://blog.decred.org/2019/08/28/Iterating-Privacy/) feature is based on the [CoinShuffle++](https://github.com/decred/cspp) protocol.
+
+## Using the Lightning Network for Privacy
+
+Privacy attacks like [common-input-ownership heuristic](https://en.bitcoin.it/wiki/Privacy#Common-input-ownership_heuristic), [address reuse](https://en.bitcoin.it/wiki/Privacy#Address_reuse) and change address detection fundamentally don’t work on any off-chain transactions technology. The Lightning Network is an off-chain transaction technology. One possible downside is that the operator of a lightning node could keep logs of the pseudonymous identifiers, the balances, and the changes of the balances of the participants. This node could later sell this information to governments or corporations. A research paper called [BOLT](https://eprint.iacr.org/2016/701) proposes solutions for this problem.
+
+
+## Hiding Cryptocurrency Usage Itself
+
+Use VPN or Tor. Don’t link your VPN or Tor usage with social networks, or with any other software or website that might identify you. Keep in mind that cookies exist. Use a browser profile that is dedicated to your crypto-related activities only. Quit all programs that you don’t need. Download some random data, so the size of your downloaded data won’t be exactly blockchain-sized. It’s a good idea to use virtual machines.
+
+## Platform-related Questions
+
+Keep in mind that the platform you are using can also have vulnerabilities. A VPN does not provide any means of defense if the system itself that you are using is hacked. Try to use open-source software, and if possible, open-source hardware as well. See [General-Security](../advanced/general-security.md) for more information.
+
+Do keep in mind that nothing is bullet-proof! If you are a person in extreme danger, for example, because you live in a territory controlled by an oppressive regime, do more research on the topic. 
+
+## More Information
+
+[Coin Control](https://medium.com/@nopara73/coin-control-is-must-learn-if-you-care-about-your-privacy-in-bitcoin-33b9a5f224a2)  
+[Merge Avoidance](https://medium.com/@octskyward/merge-avoidance-7f95a386692f)  
+[Bitcoin Wiki - Privacy](https://en.bitcoin.it/wiki/Privacy#Bad_privacy_example_-_Exchange_front_running)  
+[Bitcoin Transaction Graph Analysis (University paper)](https://arxiv.org/abs/1502.01657)
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -99,6 +99,8 @@ nav:
 - Proof-of-Work Mining:
   - 'Overview': 'mining/overview.md'
   - 'gominer Pool Mining': 'mining/proof-of-work/pool-mining/gominer.md'
+- Privacy:
+  - 'General Privacy': 'privacy/general-privacy.md'
 - FAQ:
   - 'General': 'faq/general.md'
   - 'Configuration': 'faq/configuration.md'


imestin:
This is my base idea of this. I will add "read more" links and extend the text a little bit, but this would be the general structure. Do you think this is good direction?

xaur:
I really like the general direction of thinking here, the intent to bring crypto privacy considerations into people's awareness, and the contrast with a bank account. The (biggest?) problem with privacy is people are not aware of what's going on with their data. Using simple language is a bridge that helps them get started. In crypto it is more important than earlier because money is on the line. Ultimately, this is about sovereignty too, although "sovereignty" is yet another thing that people don't often think about.

> newly generated addresses even on a website

Not clear what this refers to.

> For using a VSP, not using the same address is not possible yet.

Maybe even with solo? This needs checking.

> If we only spend fraction of an UTXO

This made me think about the target audience of this doc, or in other words, knowledge level required to read this. This bit implies the reader is familiar with what is an "output" and what is an "UTXO".

I'm not immediately suggesting to define all terms from scratch in this doc, rather to think one step higher and consider the "requirements" for this doc.

Re identification, I would add that many exchanges today voluntarily feed chain analysis companies with their customer data. This creates an interesting snowball dynamic where such company's database is enriched with every client (that must feed the data by agreement). Such databases only grow and become more and more powerful. I bet nothing is ever deleted from them.

> [General-Security](https://docs.decred.org/advanced/general-security/)

I guess you don't need a full absolute URL for local links, but dcrdocs people should know better.

imestin:
Thank you for your feedback.



>     newly generated addresses even on a website
I meant, if you have a website where people can send you money, probably you should use a script that generates new addresses every time. I will re-frame this somehow.

I will do a little bit more research about VSPs.

About UTXOs, yes, I was not really sure whether this needs explanation or not. I'm not sure indeed whether this doc should be under "Advanced". What does "Advanced" mean in the docs? What level of knowledge we expect?

s-ben:
@imestin, the Advanced section is something of a dumping ground for pages that don't fit elsewhere. Later, we may reorg the docs and come up with a better place for this, but Advanced is probably the best place for this page now. For a point of comparison, this feels similar to the [Security](https://docs.decred.org/advanced/general-security/) page.

imestin:
I changed a little bit the `newly generated addresses even on a website` part and added a link to an outside resource that explains UTXOs. I don't want to explain UTXOs inside the docs, but if you think it's necessary I can.

Soon I will add more updates.

imestin:
I think now it covers the things a basic user should know about.
I don't know if the relative links will work when it's uploaded to the website.

docs/advanced/general-privacy.md
```diff
@@ -0,0 +1,63 @@
+# General information on Privacy
+
+## About transaction graphs
```
jholdstock: Capitalise Transaction Graphs

docs/advanced/general-privacy.md
```diff
@@ -0,0 +1,63 @@
...
+
+It’s important to understand, that transactions are pseudonymous, but not anonymous. That means, a transaction graph can be built up, and there exists companies that are doing that work. If at some point one of our transactions is connected to our real world identity, it becomes easy to make assumptions based on that, for example that other transactions also belong to us, or that a specific unspent transaction belongs to us (that could give information about our balance). 
+By  checking the status of our transaction on a blockchain explorer, our IP address can be connected to one of our wallet addresses, this could create a starting point for a transaction graph. A possible solution for this is to use blockchain explorer only over Tor.
```
jholdstock: Added some extra clarification for this paragraph:

> One possible way your personal identity could be linked to your blockchain activity is through the use of a block explorer. The administrator of the block explorer may be able to view your IP address and which addresses or transactions you are browsing, and they may infer that you own those addresses or transactions. A possible solution for this is to use blockchain explorers exclusively over Tor.

docs/advanced/general-privacy.md
```diff
@@ -0,0 +1,63 @@
...
+
+## Transaction broadcasting
+
+Broadcasting the transaction can reveal that the transaction you are broadcasting belongs to you. A solution for this is to broadcast your transaction through a website, over Tor. For example, [https://explorer.dcrdata.org/decodetx](https://explorer.dcrdata.org/decodetx)
```
jholdstock: Broadcasting **a** transaction* 


also, rather than duplicating the URL using \[]\(), we can just use <>.

`<https://explorer.dcrdata.org/decodetx>`

docs/advanced/general-privacy.md
```diff
@@ -0,0 +1,63 @@
...
+It's also important to note, that different wallets might use different [address types](address-details). This can also give a hint about your identity, which combined with other signals can ultimately reveal your identity.
+
+## Trade-offs of reusing vs not reusing addresses
+
+For privacy, the best practice is to not reuse an address ever. A cryptocurrency address can be thought of as a cryptocurrency invoice instead. The term "address" does not mean, that it was designed to reuse it, like one would do with a bank account number for example. If someone would really care about privacy he or she would use a new receiving address even on a website, generated by a script.
```
jholdstock: The comma after "does not mean" is unnecessary

docs/advanced/general-privacy.md
```diff
@@ -0,0 +1,63 @@
...
+It's also important to note, that different wallets might use different [address types](address-details). This can also give a hint about your identity, which combined with other signals can ultimately reveal your identity.
+
+## Trade-offs of reusing vs not reusing addresses
+
+For privacy, the best practice is to not reuse an address ever. A cryptocurrency address can be thought of as a cryptocurrency invoice instead. The term "address" does not mean, that it was designed to reuse it, like one would do with a bank account number for example. If someone would really care about privacy he or she would use a new receiving address even on a website, generated by a script. 
+On the other hand, address reuse is preferred in situations when privacy is not important and we don't want to make the mistake of sending over large amount of funds to a wrong address. Reusing addresses is also preferred when transparency is important.
+For POS voting, not using the same address is not possible yet.
```
jholdstock: I would remove this line about PoS voting because its not strictly true. Also it will become redundant when privacy is implemented into VSPs

docs/advanced/general-privacy.md
```diff
@@ -0,0 +1,63 @@
...
+It's also important to note, that different wallets might use different [address types](address-details). This can also give a hint about your identity, which combined with other signals can ultimately reveal your identity.
+
+## Trade-offs of reusing vs not reusing addresses
+
+For privacy, the best practice is to not reuse an address ever. A cryptocurrency address can be thought of as a cryptocurrency invoice instead. The term "address" does not mean, that it was designed to reuse it, like one would do with a bank account number for example. If someone would really care about privacy he or she would use a new receiving address even on a website, generated by a script. 
+On the other hand, address reuse is preferred in situations when privacy is not important and we don't want to make the mistake of sending over large amount of funds to a wrong address. Reusing addresses is also preferred when transparency is important.
```
jholdstock: > On the other hand, there may be situations where address reuse may be preferred, or even necessary. The Decred Treasury is a good example of this. Transparency and auditability are more important than privacy in the context of the treasury.

docs/advanced/general-privacy.md
```diff
@@ -0,0 +1,63 @@
...
+For POS voting, not using the same address is not possible yet. 
+
+## Change address
+
+If we only spend fraction of an [UTXO](https://coincentral.com/utxo-beginners-explainer/), a change output will be generated. This is not good for privacy. For example, the change UTXO will be merged with another unspent transaction output next time when we are spending money, thus creating a signal that both of the transactions belong to us. An average wallet software won’t give us information about this, we won’t know, for example, that 2 UTXOs were merged to create the transaction. If you care about privacy, you should exactly know which UTXOs are being spent. [This](https://medium.com/@nopara73/coin-control-is-must-learn-if-you-care-about-your-privacy-in-bitcoin-33b9a5f224a2) is a good article on coin control.
```
jholdstock: **a** fraction*

jholdstock: Need to explain that UTXO == unspent transaction outputs. We will normally introduce an acronym like this:

> Unspent Transaction Output (UTXO)

Then we can freely use "UTXO" and the user will know what it means

docs/advanced/general-privacy.md
```diff
@@ -0,0 +1,63 @@
...
+
+## Avoid identification when acquiring cryptocurrencies 
+
+Try to avoid getting cryptocurrencies from websites that require KYC or AML, or any other means of identification. Be aware that centralized exchanges can be hacked, thus providing the collected information to the hacker, who can belong to any kind of organization. It’s also possible that at some point an oppressive government will request the collected information. ATMs can be equipped with cameras, or there can be cameras nearby the ATM. This combined with the blockchain can reveal our identity. 
+
+## Avoid leaving identifiers in public places
+
+Many times this is something physical, for example leaving behind a printed QR code somewhere, writing our nickname or real name on a piece of paper. Your name can be inprinted on the next page of the sticky notes stack, for example. Or that could also happen, that you leave behind your phone or computer unlocked for some minutes, with important information displayed on screen.
+
+## Centralized mixers
```
jholdstock: This section seems to be completely focused on bitcoin. Can we please either make this Decred-centric, or remove it.

docs/advanced/general-privacy.md
```diff
@@ -0,0 +1,63 @@
...
+Either we use a centralized mixer or the later mentioned solutions, [amount correlation](https://en.bitcoin.it/wiki/Privacy#Amount) is a problem. If we send in 1.435 BTC to a mixer, and someone receives 1.434 BTC that's a quite strong indication that the transaction belongs to us.
+Amount correlation will be a problem even if we use a so called _black box privacy technique_, like sending in our money to an anonymous exchange, exchanging it to a privacy coin which uses [zero-knowledge proof](https://en.wikipedia.org/wiki/Zero-knowledge_proof), and then switching back the money to our original currency. Amount correlation can still lead to loss of privacy. One can address this issue by breaking up the total balance to random chunks and distributing it to newly generated addresses, over a longer period of time.
+
+## Non-custodial coin mixing
+
+It is possible to mix coins without a centralized exchange. One method is [CoinJoin](https://en.bitcoin.it/wiki/CoinJoin), which will combine multiple payments from multiple spenders into a single transaction.
```
jholdstock: We should link to Decred implementations of coinjoin/coinswap. Otherwise this is not relevant.

docs/advanced/general-privacy.md
```diff
@@ -0,0 +1,63 @@
...
+Other method is [CoinSwap](https://en.bitcoin.it/wiki/CoinSwap), which is using atomic swaps. A third method is [TumbleBit](https://en.bitcoin.it/wiki/TumbleBit), which allows parties to make payments through an untrusted Tumbler.
+[Decred's build in coin mixing](https://blog.decred.org/2019/08/28/Iterating-Privacy/) feature is based on the CoinShuffle++ protocol.
+
+## Using the Lightning Network for privacy
+
+Privacy attacks like [common-input-ownership heuristic](https://en.bitcoin.it/wiki/Privacy#Common-input-ownership_heuristic), [address reuse](https://en.bitcoin.it/wiki/Privacy#Address_reuse) and change address detection fundamentally don’t work on any off-chain transactions technology. The Lightning Network is an off-chain transaction technology. One possible downside is that the operator of the lightning node could keep logs.
```
jholdstock: As a reader, I am wondering what information can a lightning node log, and why is that a privacy risk?

docs/advanced/general-privacy.md
```diff
@@ -0,0 +1,63 @@
...
+
+
+## Hiding cryptocurrency usage itself
+
+Use VPN or Tor. Don’t link your VPN or Tor usage with social networks, or with any other software or website that might identify you. Keep in mind that cookies exist. Use a browser profile that is dedicated to your crypto-related activities only. Quit all programs that you don’t need. Download some random data, so the size of your downloaded data won’t be exactly blockchain-sized. It’s a good idea to use virtual machines.
+
+## Platform-related questions
+
+Keep in mind that the platform you are using can also have vulnerabilities. A VPN does not provide any means of defense if the system itself that you are using is hacked. Try to use open-source software, and if possible, open-source hardware as well. See [General-Security](general-security) for more information.
```
jholdstock: What do you mean by platform here? Operating system?

Link to `general-security.md` with markdown file extension.

docs/advanced/general-privacy.md
imestin: The reason I included it is that I wanted the reader to think about what kind of things exist in this field. I didn't wanted him/her to start using these tools, just to start thinking about how privacy tech can work.

docs/advanced/general-privacy.md
imestin: I included these paragraph to make the reader think about privacy tech. I was not trying to direct the reader to start using these tools, I included them for historic reasons mainly. I know that it is a bit focused on bitcoin, but I don't know how otherwise to make the reader think about the basic concept.

I can try to re-frame this paragraph, but the basic intention is not to give present-day information in this section. CoinShuffle++ was meant to be present day information.

docs/advanced/general-privacy.md
imestin: https://cspp.decred.org/

I found this, but as I know, CoinShuffle++ and CoinJoin is not exactly the same. I wouldn't link to this resource.

What do you think?

xaur:
I suggest to use web page links instead of PDF links for the paper. This way the reader can click and see a summary in his browser before downloading the PDF

https://arxiv.org/abs/1502.01657
https://eprint.iacr.org/2016/701

and so on

imestin:
@xaur but you can not actually read the research paper when you click on these links. For reading you will have to download the PDF, or you can open the PDF with your browser, if your browser supports this.

On the other hand, this way the user could see what he is about to download, before actually downloading.

I wouldn't change it.

imestin:
what do you think about the issue with the centralized mixers? Do you think it's good idea to provide information which is historical, or would you only provide Decred-specific info?

xaur:
> but you can not actually read the research paper when you click on these links. For reading you will have to download the PDF

It is trivial to find the PDF download on those pages, but it is not trivial to go in the opposite direction, i.e. to change the PDF link into the abstract link. I think it's useful to read the abstract before downloading the paper.

re centralized mixers, I think the Decred-focused page should only include information that is helpful to understand Decred's implementation. In that context, mentioning existing mixer tech (even for Bitcoin) is ok IMO. However, any significant exploration of non-Decred mixing tech should live in a separate page about general privacy in crypto.

docs/advanced/general-privacy.md
```diff
@@ -0,0 +1,62 @@
...
+Amount correlation will be a problem even if we use a so called _black box privacy technique_, like sending in our money to an anonymous exchange, exchanging it to a privacy coin which uses [zero-knowledge proof](https://en.wikipedia.org/wiki/Zero-knowledge_proof), and then switching back the money to our original currency. Amount correlation can still lead to loss of privacy. One can address this issue by breaking up the total balance to random chunks and distributing it to newly generated addresses, over a longer period of time.
+
+## Non-custodial Coin Mixing
+
+It is possible to mix coins without a centralized exchange. One method is [CoinJoin](https://en.bitcoin.it/wiki/CoinJoin), which will combine multiple payments from multiple spenders into a single transaction.
+Other method is [CoinSwap](https://en.bitcoin.it/wiki/CoinSwap), which is using atomic swaps. A third method is [TumbleBit](https://en.bitcoin.it/wiki/TumbleBit), which allows parties to make payments through an untrusted Tumbler.
```
dajohi: could reference github.com/decred/tumblebit and github.com/decred/cspp around here...

docs/advanced/general-privacy.md
imestin: I shortened the Centralized Mixers section

docs/advanced/general-privacy.md
imestin: I removed CoinJoin and CoinSwap

imestin:
I will create privacy folder, as suggested in the post from @xaur 

> There are two ways to organize this.

First is group by aspect: create new Privacy group as suggested above and let it host both general privacy and Decred privacy. This is similar to existing Governance group.

The other way I didn't imagine initially is to fit into existing grouping by use case. Place the general privacy page under Advanced (next to Security), and then place another page for Decred privacy features in Wallets category, similar to how it hosts a page for Lightning Network use case.

xaur:
@imestin please not that I'm just lurking around while @jholdstock and @s-ben actively maintain this repo. I hope they follow this issue to prevent wasted work.

docs/advanced/general-privacy.md
```diff
@@ -0,0 +1,61 @@
...
+
+## Avoid Identification when Acquiring Cryptocurrencies 
+
+Try to avoid getting cryptocurrencies from websites that require KYC or AML, or any other means of identification. Be aware that centralized exchanges can be hacked, thus providing the collected information to the hacker, who can belong to any kind of organization. It’s also possible that at some point an oppressive government will request the collected information. ATMs can be equipped with cameras, or there can be cameras nearby the ATM. This combined with the blockchain can reveal our identity. 
+
+## Avoid Leaving Identifiers in Public Places
+
+Many times this is something physical, for example leaving behind a printed QR code somewhere, writing our nickname or real name on a piece of paper. Your name can be inprinted on the next page of the sticky notes stack, for example. Or that could also happen, that you leave behind your phone or computer unlocked for some minutes, with important information displayed on screen.
```
gozart1: I noticed the word "our" and "your" is used interchangeably throughout this file. For example it says "our nickname" here and that just seems off to me. Suggest just sticking to "your" in all cases.

PR #989, opened 2019-09-18
Adds full node setup page

### LINKED ISSUE #987: Add page full node setup
Per this [conversation](https://matrix.to/#/!tfqymymiNgzSUJTHqS:decred.org/$156820079627471eFYxw:decred.org?via=decred.org&via=matrix.org&via=zettaport.com), it would be good to have a page with instructions for setting up a full node. @noahpierau has provided the below draft,
https://github.com/noahpierau/articles/blob/master/how-to-run-a-decred-full-node.md

Notes on draft from chat:

jholdstock
>step 7: would be better to put the externalip into the dcrd config file rather than providing it through the flag every time dcrd is started up
>Also you may want to add a note about external ip address needing to remain static. This is quite important for your specific target audience because most household internet connections dont allow this

davecgh
>1.5.0 will no longer need extenalip

Hanon
>yes that's a great improvement! will 1.5.0 automatically detect ip changes as well?

davecgh
>Sort of, but it would take a long time because it keeps a score

Note that 1.5 is an upcoming release.

s-ben:
Adds a page with instructions for setting up a Decred full node ( Closes #987 ). 

Have used @noahpierau's [draft](https://github.com/noahpierau/articles/blob/master/how-to-run-a-decred-full-node.md) as a baseline, adding in material from some other guides I found in doing my research:

- _Checkmatey's guide (@checkmatey): https://medium.com/@_Checkmatey_/running-a-decred-raspberry-pi-node-ac605b70c652 
- Stakey.club (anyone know their GitHub?):: https://stakey.club/en/installing-dcrd/
- Pablito's guide (anyone know his GitHub?): https://medium.com/crypto-rocket-blog/my-experience-setting-up-a-decred-full-node-8a9bbf55bc30

I know the external IP address will no longer be needed in v1.5.0, but don't want to include anything about that until the release.

### DIFF
--- a/docs/advanced/full-node.md
+++ b/docs/advanced/full-node.md
@@ -38,7 +38,7 @@ If running a node on your own hardware (i.e. not on remote servers), you will al
 
 Below are some guides from the community that walk you step-by-step through setting up a full node on various hardware and operating systems. For generic high-level instructions, see the [Steps](#steps) section below. 
 
-- [Running a Decred Raspberry Pi Node](https://medium.com/@_Checkmatey_/running-a-decred-raspberry-pi-node-ac605b70c652): A more beginner-friedly guide that uses a VNC server (remote desktop). This guide is intented to be a gentle introduction to using SSH on the command line. 
+- [Running a Decred Raspberry Pi Node](https://medium.com/@_Checkmatey_/running-a-decred-raspberry-pi-node-ac605b70c652): A more beginner-friendly guide that uses a VNC server (remote desktop). This guide is intented to be a gentle introduction to using SSH on the command line. 
 - [My experience setting up a Decred full-node](https://medium.com/crypto-rocket-blog/my-experience-setting-up-a-decred-full-node-8a9bbf55bc30): Another beginner-friendly guide that walks through setting up a full node on a Raspberry Pi using an Ubuntu terminal on a Windows PC.
 - [Installing dcrd](https://stakey.club/en/installing-dcrd/): A more advanced guide to installing a full node (Mac or Debian) that also covers manual installation (from binaries or source) and configuring dcrd to run on Tor for extra security.
 
--- a/docs/advanced/full-node.md
+++ b/docs/advanced/full-node.md
@@ -61,7 +61,7 @@ Below are the basic high-level steps to install a full node on any hardware and 
 1. **Restart dcrd**
 	- Restart `dcrd` and pass it your external IP address using the following flag `--externalip=[your.external.ip]`
 	- If you don't wish to pass your IP address every time you restart `dcrd`, you can put it in dcrd's [configuration file](../wallets/cli/startup-basics.md#configuration-file-locations). \
-	- If you have configured your node to allow inbound peers, you should see inbound peers in the logs. Inbound peers can also be check using the Remote Procedure Call (RPC) command [getpeerinfo](../wallets/cli/dcrctl-rpc-commands.md).
+	- If you have configured your node to allow inbound peers, you should see inbound peers in the logs. Inbound peers can also be checked using the Remote Procedure Call (RPC) command [getpeerinfo](../wallets/cli/dcrctl-rpc-commands.md).
 1. **Leave running**
 	-  Leave your node running, online, in a safe space, 24/7/365.
 
--- a/docs/advanced/full-node.md
+++ b/docs/advanced/full-node.md
@@ -60,7 +60,8 @@ Below are the basic high-level steps to install a full node on any hardware and 
 	- Check a Decred block explorer such as [dcrdata](https://dcrdata.decred.org/) and wait until the block height displayed matches the block height shown locally.
 1. **Restart dcrd**
 	- Restart `dcrd` and pass it your external IP address using the following flag `--externalip=[your.external.ip]`
-	- If you don't wish to pass your IP address every time you restart `dcrd`, you can put it in dcrd's [configuration file](../wallets/cli/startup-basics.md#configuration-file-locations). 
+	- If you don't wish to pass your IP address every time you restart `dcrd`, you can put it in dcrd's [configuration file](../wallets/cli/startup-basics.md#configuration-file-locations). \
+	- If you have configured your node to allow inbound peers, you should see inbound peers in the logs. Inbound peers can also be check using the Remote Procedure Call (RPC) command [getpeerinfo](../wallets/cli/dcrctl-rpc-commands.md).
 1. **Leave running**
 	-  Leave your node running, online, in a safe space, 24/7/365.
 
--- a/docs/advanced/full-node.md
+++ b/docs/advanced/full-node.md
@@ -1,0 +1,75 @@
+# <img class="dcr-icon" src="/img/dcr-icons/AtoB.svg" /> Running a Full Node
+
+---
+
+This guide explains everything you need to know about running a full node on Decred.
+
+## What is a full node?
+
+A full node is a program that fully validates transactions and blocks, without having to rely on a third party. For this guide, we'll be using [dcrd](https://github.com/decred/dcrd), the most popular full node implementation for Decred. 
+
+## Why run a full node?
+
+There are a number of reasons to run a full node:
+
+- **Help the network:** Full nodes are the backbone of Decred. By validating and broadcasting transactions without relying on third parties&mdash; as lightweight nodes do&mdash;they help keep the network decentralized. Each full node added to the network adds to its security and robustness. 
+- **Safety & Privacy:** Because a lightweight node cannot fully validate transactions and blocks, it must trust third parties to do so. This means that the third party will have access to information about your transactions, as well as any personal data you give to that third party. For instance, many lightweight wallets rely on a centralized server to process transactions. Even Simplified Payment Verification (SPV) wallets, which typically connect directly to full nodes in the network, still rely on those other full nodes to process and broadcast their transactions. 
+
+
+## Requirements:
+
+You do not need high-performance hardware to run a full node. A cheap (refurbished) laptop or Rasberry Pi (3B+ or better) will work. Below are the basic requirements:
+
+- \>15 GB disk space (as of September 2019, increases over time)
+- \>2 GB of RAM
+- Stable Operating System (OS) &mdash;Windows 7/8.x/10 (server preferred), macOS, Linux
+- Reliable internet connection
+
+If running a node on your own hardware (i.e. not on remote servers), you will also need:
+
+- Access to the settings of your router (to allow outbound connections)
+- A safe place to leave your device running
+- If you need a silent node (hard disk drives can be loud), a SSD or USB drive
+
+
+
+## Guides
+
+Below are some guides from the community that walk you step-by-step through setting up a full node on various hardware and OSes. For generic high-level instructions, see the [Steps](#steps) section below. 
+
+- [Running a Decred Raspberry Pi Node](https://medium.com/@_Checkmatey_/running-a-decred-raspberry-pi-node-ac605b70c652): A more beginner friedly guide that uses a VNC server (remote desktop). This guide is intented to be a gentle introduction to using SSH on the command line. 
+- [My experience setting up a Decred full-node](https://medium.com/crypto-rocket-blog/my-experience-setting-up-a-decred-full-node-8a9bbf55bc30): Another beginner-friendly guide that walks through setting up a full node on a Raspberry Pi using an Ubuntu terminal on a Windows PC.
+- [Installing dcrd](https://stakey.club/en/installing-dcrd/): A more advanced guide to installing a full node (Mac or Debian) that also covers manual installation (from binaries or source) and configuring dcrd to run on Tor for extra security.
+
+## Steps
+
+Below are the basic high-level steps to install a full node on any hardware and OS. These instructions use [dcrinstall](../wallets/cli/cli-installation.md), the automatic installer and upgrader software. In addition to installing `dcrd`, it will also install `dcrwallet` (which allows you to create wallets) and `dcrctl` (which allows you to control `dcrwallet` and `dcrd` from the command line). You can also just install `dcrd` on its own from [binaries or source files](https://github.com/decred/dcrd#installing-and-updating); a full node does not require a wallet unless you want to use your full node to validate and broadcast your own transactions.
+
+
+1. **Download and install dcrd**
+	- Follow the instructions for your OS in the [CLI Installation guide](https://docs.decred.org/wallets/cli/cli-installation/). The `dcrd` executable will be installed in a directory named `/decred` under your OS's home directory. 
+2. **Start `dcrd`**
+	- Navigate to the `/decred` directory and launch `dcrd`. See the [Operating System Differences](../wallets/cli/os-differences.md) page for OS-specific commands.
+	- `dcrd` will boot up, begin connecting to peers and downloading the full Decred blockchain.
+3. **(Optional) Create wallet**
+	- If you would like to use your full node to validate and broadcast your wallet's transactions, you'll need to create and set up a wallet. See the [dcrwallet Setup Guide](../wallets/cli/dcrwallet-setup.md) for instructions.
+4. **Enable incoming connections**
+	- `dcrd` will automatically begin downloading the blockchain and connecting to peers. However, if you want to help the network and have your node start validating transactions, you'll need to configure your internet router to allow inbound connections. 
+		1. Find your [Local IP address](https://www.howtogeek.com/236838/how-to-find-any-devices-ip-address-mac-address-and-other-network-connection-details/) and [External IP address](https://whatismyipaddress.com/).
+		2. Edit the settings of your router to open port 9108 (port forwarding). This [process](https://www.wikihow.com/Set-Up-Port-Forwarding-on-a-Router) depends on the type of router you have. 
+	- If running a firewall, you'll also need to configure it to allow connections on port 9108.
+5. **Wait for dcrd to sync**
+	- Check a Decred block explorer such as [dcrdata](https://dcrdata.decred.org/) and wait until the block height displayed matches the block height shown locally.
+5. **Restart dcrd**
+	- Restart `dcrd` and pass it your external IP address using the following flag `--externalip=[your.external.ip]`
+	- If you don't wish to pass your IP address every time you restart `dcrd`, you can put it in dcrd's [configuration file](../wallets/cli/startup-basics.md#configuration-file-locations). 
+
+Congratulations, you are now running a Decred full node to support the network!
+
+
+
+
+
+
+
+
--- a/docs/advanced/full-node.md
+++ b/docs/advanced/full-node.md
@@ -2,7 +2,7 @@
 
 ---
 
-This guide explains everything you need to know about running a full node on Decred.
+This guide explains everything you need to know about running a Decred full node.
 
 ## What is a full node?
 
@@ -12,29 +12,33 @@ A full node is a program that fully validates transactions and blocks, without h
 
 There are a number of reasons to run a full node:
 
-- **Help the network:** Full nodes are the backbone of Decred. By validating and broadcasting transactions without relying on third parties&mdash; as lightweight nodes do&mdash;they help keep the network decentralized. Each full node added to the network adds to its security and robustness. 
-- **Safety & Privacy:** Because a lightweight node cannot fully validate transactions and blocks, it must trust third parties to do so. This means that the third party will have access to information about your transactions, as well as any personal data you give to that third party. For instance, many lightweight wallets rely on a centralized server to process transactions. Even Simplified Payment Verification (SPV) wallets, which typically connect directly to full nodes in the network, still rely on those other full nodes to process and broadcast their transactions. 
+- **Help the network:** Full nodes are the backbone of Decred. By validating and broadcasting transactions without relying on third parties they help keep the network decentralized. Each full node added to the network adds to its security and robustness. 
+- **Safety & Privacy:** Because a lightweight node cannot fully validate transactions and blocks, it must trust third parties to do so. This means that the third party will have access to information about your transactions, as well as any personal data you give to that third party. For instance, many lightweight wallets rely on a centralized server to process transactions. Even [Simplified Payment Verification (SPV)](../wallets/spv.md) wallets, which typically connect directly to full nodes in the network, still rely on those other full nodes to process and broadcast their transactions. 
 
 ## Requirements:
 
 Below are the basic requirements for running a full node:
 
-- \>15 GB disk space (as of September 2019, increases over time)
+- \>15 GB disk space (as of September 2019). 
+
+!!! note "Note"
+    The size of the blockchain is continually increasing as new blocks are added to the chain. To support migration to new versions and future growth, a good rule of thumb is to have ~3X the [current raw blockchain size](https://dcrdata.decred.org/charts?chart=blockchain-size&zoom=ikd7pc00-k0vn9c00&bin=day&axis=time) in hard disk space.  
+
 - \>2 GB of RAM
-- Stable Operating System (OS) &mdash;Windows 7/8.x/10 (server preferred), macOS, Linux
+- Supported Operating System (OS) &mdash;Windows 7/8.x/10 (server preferred), macOS, Linux
 - Reliable internet connection
 
 If running a node on your own hardware (i.e. not on remote servers), you will also need:
 
-- Access to the settings of your router (to allow outbound connections)
+- Ability to open a port for incoming internet traffic. If using a home wifi network, this may require changing the settings of your router.
 - A safe place to leave your device running
 - If you need a silent node (hard disk drives can be loud), a SSD or USB drive
 
 ## Guides
 
-Below are some guides from the community that walk you step-by-step through setting up a full node on various hardware and OSes. For generic high-level instructions, see the [Steps](#steps) section below. 
+Below are some guides from the community that walk you step-by-step through setting up a full node on various hardware and operating systems. For generic high-level instructions, see the [Steps](#steps) section below. 
 
-- [Running a Decred Raspberry Pi Node](https://medium.com/@_Checkmatey_/running-a-decred-raspberry-pi-node-ac605b70c652): A more beginner friedly guide that uses a VNC server (remote desktop). This guide is intented to be a gentle introduction to using SSH on the command line. 
+- [Running a Decred Raspberry Pi Node](https://medium.com/@_Checkmatey_/running-a-decred-raspberry-pi-node-ac605b70c652): A more beginner-friedly guide that uses a VNC server (remote desktop). This guide is intented to be a gentle introduction to using SSH on the command line. 
 - [My experience setting up a Decred full-node](https://medium.com/crypto-rocket-blog/my-experience-setting-up-a-decred-full-node-8a9bbf55bc30): Another beginner-friendly guide that walks through setting up a full node on a Raspberry Pi using an Ubuntu terminal on a Windows PC.
 - [Installing dcrd](https://stakey.club/en/installing-dcrd/): A more advanced guide to installing a full node (Mac or Debian) that also covers manual installation (from binaries or source) and configuring dcrd to run on Tor for extra security.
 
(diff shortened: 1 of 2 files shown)


RichardRed0x:
I'm planning to give this a thorough review when I finish what I'm working on. In a day or two I guess.

docs/advanced/full-node.md
```diff
@@ -0,0 +1,75 @@
...
+
+There are a number of reasons to run a full node:
+
+- **Help the network:** Full nodes are the backbone of Decred. By validating and broadcasting transactions without relying on third parties&mdash; as lightweight nodes do&mdash;they help keep the network decentralized. Each full node added to the network adds to its security and robustness. 
+- **Safety & Privacy:** Because a lightweight node cannot fully validate transactions and blocks, it must trust third parties to do so. This means that the third party will have access to information about your transactions, as well as any personal data you give to that third party. For instance, many lightweight wallets rely on a centralized server to process transactions. Even Simplified Payment Verification (SPV) wallets, which typically connect directly to full nodes in the network, still rely on those other full nodes to process and broadcast their transactions. 
+
+
+## Requirements:
+
+You do not need high-performance hardware to run a full node. A cheap (refurbished) laptop or Rasberry Pi (3B+ or better) will work. Below are the basic requirements:
```
davecgh: I'd be careful about putting any claims to this effect in the documentation since, while it's currently true, as the blockchain and usage grows, it will likely no longer be true.

docs/advanced/full-node.md
```diff
@@ -0,0 +1,75 @@
...
+A full node is a program that fully validates transactions and blocks, without having to rely on a third party. For this guide, we'll be using [dcrd](https://github.com/decred/dcrd), the most popular full node implementation for Decred. 
+
+## Why run a full node?
+
+There are a number of reasons to run a full node:
+
+- **Help the network:** Full nodes are the backbone of Decred. By validating and broadcasting transactions without relying on third parties&mdash; as lightweight nodes do&mdash;they help keep the network decentralized. Each full node added to the network adds to its security and robustness. 
+- **Safety & Privacy:** Because a lightweight node cannot fully validate transactions and blocks, it must trust third parties to do so. This means that the third party will have access to information about your transactions, as well as any personal data you give to that third party. For instance, many lightweight wallets rely on a centralized server to process transactions. Even Simplified Payment Verification (SPV) wallets, which typically connect directly to full nodes in the network, still rely on those other full nodes to process and broadcast their transactions. 
+
+
```
davecgh: Inconsistent number of blank lines here and elsewhere.  I see a mix of 1, 2, and 3 blank lines.

docs/advanced/full-node.md
s-ben: Good point. In the next section, we have a couple guides which use a Raspberri Pi, which should convey that possibility.

docs/advanced/full-node.md
s-ben: Fixed

docs/advanced/full-node.md
```diff
@@ -0,0 +1,63 @@
...
+- [Installing dcrd](https://stakey.club/en/installing-dcrd/): A more advanced guide to installing a full node (Mac or Debian) that also covers manual installation (from binaries or source) and configuring dcrd to run on Tor for extra security.
+
+## Steps
+
+Below are the basic high-level steps to install a full node on any hardware and OS. These instructions use [dcrinstall](../wallets/cli/cli-installation.md), the automatic installer and upgrader software. In addition to installing `dcrd`, it will also install `dcrwallet` (which allows you to create wallets) and `dcrctl` (which allows you to control `dcrwallet` and `dcrd` from the command line). You can also just install `dcrd` on its own from [binaries or source files](https://github.com/decred/dcrd#installing-and-updating); a full node does not require a wallet unless you want to use your full node to validate and broadcast your own transactions.
+
+1. **Download and install dcrd**
```
jholdstock: Numbered list should use `1)` on every line

docs/advanced/full-node.md
```diff
@@ -0,0 +1,63 @@
...
+	- `dcrd` will boot up, begin connecting to peers and downloading the full Decred blockchain.
+3. **(Optional) Create wallet**
+	- If you would like to use your full node to validate and broadcast your wallet's transactions, you'll need to create and set up a wallet. See the [dcrwallet Setup Guide](../wallets/cli/dcrwallet-setup.md) for instructions.
+4. **Enable incoming connections**
+	- `dcrd` will automatically begin downloading the blockchain and connecting to peers. However, if you want to help the network and have your node start validating transactions, you'll need to configure your internet router to allow inbound connections. 
+		1. Find your [Local IP address](https://www.howtogeek.com/236838/how-to-find-any-devices-ip-address-mac-address-and-other-network-connection-details/) and [External IP address](https://whatismyipaddress.com/).
+		2. Edit the settings of your router to open port 9108 (port forwarding). This [process](https://www.wikihow.com/Set-Up-Port-Forwarding-on-a-Router) depends on the type of router you have. 
+	- If running a firewall, you'll also need to configure it to allow connections on port 9108.
```
jholdstock: **inbound** connections

docs/advanced/full-node.md
```diff
@@ -0,0 +1,63 @@
...
+5. **Wait for dcrd to sync**
+	- Check a Decred block explorer such as [dcrdata](https://dcrdata.decred.org/) and wait until the block height displayed matches the block height shown locally.
+5. **Restart dcrd**
+	- Restart `dcrd` and pass it your external IP address using the following flag `--externalip=[your.external.ip]`
+	- If you don't wish to pass your IP address every time you restart `dcrd`, you can put it in dcrd's [configuration file](../wallets/cli/startup-basics.md#configuration-file-locations). 
+
```
jholdstock: Last step - leave your node running, online, in a safe place, 24/7/365

docs/advanced/full-node.md
```diff
@@ -0,0 +1,63 @@
...
+4. **Enable incoming connections**
+	- `dcrd` will automatically begin downloading the blockchain and connecting to peers. However, if you want to help the network and have your node start validating transactions, you'll need to configure your internet router to allow inbound connections. 
+		1. Find your [Local IP address](https://www.howtogeek.com/236838/how-to-find-any-devices-ip-address-mac-address-and-other-network-connection-details/) and [External IP address](https://whatismyipaddress.com/).
+		2. Edit the settings of your router to open port 9108 (port forwarding). This [process](https://www.wikihow.com/Set-Up-Port-Forwarding-on-a-Router) depends on the type of router you have. 
+	- If running a firewall, you'll also need to configure it to allow connections on port 9108.
+5. **Wait for dcrd to sync**
+	- Check a Decred block explorer such as [dcrdata](https://dcrdata.decred.org/) and wait until the block height displayed matches the block height shown locally.
+5. **Restart dcrd**
+	- Restart `dcrd` and pass it your external IP address using the following flag `--externalip=[your.external.ip]`
```
jholdstock: I think it makes more sense to set your external IP address **before** starting dcrd for the first time.

Nodes maintain a list of accessible/non-accessible peers, and coming online whilst not accessible may result in your node being added to the non-accessible lists, leading to peers ignoring your node in future.

@davecgh is my understanding correct here?

docs/advanced/full-node.md
```diff
@@ -0,0 +1,63 @@
+# <img class="dcr-icon" src="/img/dcr-icons/AtoB.svg" /> Running a Full Node
+
+---
+
+This guide explains everything you need to know about running a full node on Decred.
```
RichardRed0x: "running a Decred full node"?

docs/advanced/full-node.md
```diff
@@ -0,0 +1,63 @@
...
+
+## What is a full node?
+
+A full node is a program that fully validates transactions and blocks, without having to rely on a third party. For this guide, we'll be using [dcrd](https://github.com/decred/dcrd), the most popular full node implementation for Decred. 
+
+## Why run a full node?
+
+There are a number of reasons to run a full node:
+
+- **Help the network:** Full nodes are the backbone of Decred. By validating and broadcasting transactions without relying on third parties&mdash; as lightweight nodes do&mdash;they help keep the network decentralized. Each full node added to the network adds to its security and robustness.
```
RichardRed0x: "&mdash; as lightweight nodes do&mdash;" seems unnecessary here.

docs/advanced/full-node.md
```diff
@@ -0,0 +1,63 @@
...
+
+1. **Download and install dcrd**
+	- Follow the instructions for your OS in the [CLI Installation guide](https://docs.decred.org/wallets/cli/cli-installation/). The `dcrd` executable will be installed in a directory named `/decred` under your OS's home directory. 
+2. **Start `dcrd`**
+	- Navigate to the `/decred` directory and launch `dcrd`. See the [Operating System Differences](../wallets/cli/os-differences.md) page for OS-specific commands.
+	- `dcrd` will boot up, begin connecting to peers and downloading the full Decred blockchain.
+3. **(Optional) Create wallet**
+	- If you would like to use your full node to validate and broadcast your wallet's transactions, you'll need to create and set up a wallet. See the [dcrwallet Setup Guide](../wallets/cli/dcrwallet-setup.md) for instructions.
```
RichardRed0x: I believe dcrinstall runs through wallet creation by default, so if step 1 was followed wallet creation would be encountered there..

docs/advanced/full-node.md
```diff
@@ -0,0 +1,63 @@
...
+4. **Enable incoming connections**
+	- `dcrd` will automatically begin downloading the blockchain and connecting to peers. However, if you want to help the network and have your node start validating transactions, you'll need to configure your internet router to allow inbound connections. 
+		1. Find your [Local IP address](https://www.howtogeek.com/236838/how-to-find-any-devices-ip-address-mac-address-and-other-network-connection-details/) and [External IP address](https://whatismyipaddress.com/).
+		2. Edit the settings of your router to open port 9108 (port forwarding). This [process](https://www.wikihow.com/Set-Up-Port-Forwarding-on-a-Router) depends on the type of router you have. 
+	- If running a firewall, you'll also need to configure it to allow connections on port 9108.
+5. **Wait for dcrd to sync**
+	- Check a Decred block explorer such as [dcrdata](https://dcrdata.decred.org/) and wait until the block height displayed matches the block height shown locally.
```
RichardRed0x: Checking dcrdata seems unlikely to be necessary, dcrd keeps a running report of the data of the latest block processed every 10s, for example.

docs/advanced/full-node.md
```diff
@@ -0,0 +1,63 @@
...
+There are a number of reasons to run a full node:
+
+- **Help the network:** Full nodes are the backbone of Decred. By validating and broadcasting transactions without relying on third parties&mdash; as lightweight nodes do&mdash;they help keep the network decentralized. Each full node added to the network adds to its security and robustness. 
+- **Safety & Privacy:** Because a lightweight node cannot fully validate transactions and blocks, it must trust third parties to do so. This means that the third party will have access to information about your transactions, as well as any personal data you give to that third party. For instance, many lightweight wallets rely on a centralized server to process transactions. Even Simplified Payment Verification (SPV) wallets, which typically connect directly to full nodes in the network, still rely on those other full nodes to process and broadcast their transactions. 
+
+## Requirements:
+
+Below are the basic requirements for running a full node:
+
+- \>15 GB disk space (as of September 2019, increases over time)
```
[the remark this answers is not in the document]
davecgh: Linking to a blockchain size chart might be a reasonable idea, but it would be far more misleading without additional information than this value (which is intentionally high as discussed in a moment).  The problem is that the actual space is quite a bit more than the raw sum of the blocks.  There is a lot of metadata in addition to the raw blocks such as the spend journal, the utxo set, the upcoming required version 2 block filters, and optionally more depending on what indexes the user has chosen to enable.

Also, the value provided needs to be at least 2x whatever the current usage is in general to better support migrations and future growth, because remember that it gets larger every day, so if you actually recommend the current space, they're already going to be about of disk space in short order.  That is why it currently recommends around 3x the current raw blockchain size.

Given all of that, if we want a more dynamic approach it will need to say something like "3 times the current raw size reported at [blockchain size chart](https://dcrdata.decred.org/charts?chart=blockchain-size)".  That would obviously need to be worded better, but you get the idea.

docs/advanced/full-node.md
jholdstock: Yeah this is correct. `dcrwallet --create` is triggered by dcrinstall.

docs/advanced/full-node.md
```diff
@@ -0,0 +1,63 @@
...
+- \>2 GB of RAM
+- Stable Operating System (OS) &mdash;Windows 7/8.x/10 (server preferred), macOS, Linux
```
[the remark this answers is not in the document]
s-ben: Wasn't sure about this line actually. Do we need it at all? Mac, Windows, Linux kind of covers nearly all OSes really, so why bother listing them?

docs/advanced/full-node.md
```diff
@@ -0,0 +1,63 @@
...
+
+1. **Download and install dcrd**
+	- Follow the instructions for your OS in the [CLI Installation guide](https://docs.decred.org/wallets/cli/cli-installation/). The `dcrd` executable will be installed in a directory named `/decred` under your OS's home directory. 
+2. **Start `dcrd`**
+	- Navigate to the `/decred` directory and launch `dcrd`. See the [Operating System Differences](../wallets/cli/os-differences.md) page for OS-specific commands.
+	- `dcrd` will boot up, begin connecting to peers and downloading the full Decred blockchain.
+3. **(Optional) Create wallet**
+	- If you would like to use your full node to validate and broadcast your wallet's transactions, you'll need to create and set up a wallet. See the [dcrwallet Setup Guide](../wallets/cli/dcrwallet-setup.md) for instructions.
+4. **Enable incoming connections**
+	- `dcrd` will automatically begin downloading the blockchain and connecting to peers. However, if you want to help the network and have your node start validating transactions, you'll need to configure your internet router to allow inbound connections.
```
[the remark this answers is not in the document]
s-ben: Right, but how would it get new transactions to validate if it isn't accepting inbound connections?

docs/advanced/full-node.md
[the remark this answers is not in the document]
jholdstock: Agree, its quite pointless.

We talk about OS choices in this page https://docs.decred.org/advanced/general-security/
maybe a link there? I dont mind either way

docs/advanced/full-node.md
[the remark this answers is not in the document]
jholdstock: Inbound or outbound only describes how a connection is established, it does not describe the flow of information. Broadly speaking, an outbound connection is one you have initiated yourself by dialing out to a peer, and an inbound connection is coming in to you from another peer. 

Once a connection is established (either inbound or outbound) it enables bidirectional communication. This is how dcrd is able to participate and fully validate, even if it only has outbound peers. Inbound peers aren't strictly necessary at all, but for a node to actually have a beneficial effect on the network you should really let them in.

docs/advanced/full-node.md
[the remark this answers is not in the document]
s-ben: Right now have just stripped out the 'stable' part:

>- Operating System (OS) &mdash;Windows 7/8.x/10 (server preferred), macOS, Linux

But maybe we can just say 'Supported Operating System (OS)?'. Will dcrinstall work on any OS? This line might not be very relevant in this context, but I wonder if readers will want to see something about OS requirements anyway, just because they're use to seeing it?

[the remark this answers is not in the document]
jholdstock: Yes dcrinstall works on mac/win/linux. I dont really mind tbh, either listing them or just writing "supported operating systems" is fine

docs/advanced/full-node.md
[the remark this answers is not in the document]
s-ben: Updated to this:

>- \>15 GB disk space (as of September 2019). 

>!!! note "Note"
    The size of the blockchain is continually increasing as new blocks are added to the chain. To support migration to new versions and future growth, a good rule of thumb is to have ~3X the [current raw blockchain size](https://dcrdata.decred.org/charts?chart=blockchain-size&zoom=ikd7pc00-k0vn9c00&bin=day&axis=time) in hard disk space.

docs/advanced/full-node.md
```diff
@@ -0,0 +1,63 @@
...
+
+A full node is a program that fully validates transactions and blocks, without having to rely on a third party. For this guide, we'll be using [dcrd](https://github.com/decred/dcrd), the most popular full node implementation for Decred. 
+
+## Why run a full node?
+
+There are a number of reasons to run a full node:
+
+- **Help the network:** Full nodes are the backbone of Decred. By validating and broadcasting transactions without relying on third parties&mdash; as lightweight nodes do&mdash;they help keep the network decentralized. Each full node added to the network adds to its security and robustness. 
+- **Safety & Privacy:** Because a lightweight node cannot fully validate transactions and blocks, it must trust third parties to do so. This means that the third party will have access to information about your transactions, as well as any personal data you give to that third party. For instance, many lightweight wallets rely on a centralized server to process transactions. Even Simplified Payment Verification (SPV) wallets, which typically connect directly to full nodes in the network, still rely on those other full nodes to process and broadcast their transactions.
```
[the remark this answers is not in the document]
s-ben: Done

docs/advanced/full-node.md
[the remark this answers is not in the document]
s-ben: Have this now:

>- Supported Operating System (OS) &mdash;Windows 7/8.x/10 (server preferred), macOS, Linux

docs/advanced/full-node.md
```diff
@@ -0,0 +1,63 @@
...
+Below are the basic requirements for running a full node:
+
+- \>15 GB disk space (as of September 2019, increases over time)
+- \>2 GB of RAM
+- Stable Operating System (OS) &mdash;Windows 7/8.x/10 (server preferred), macOS, Linux
+- Reliable internet connection
+
+If running a node on your own hardware (i.e. not on remote servers), you will also need:
+
+- Access to the settings of your router (to allow outbound connections)
```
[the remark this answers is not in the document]
s-ben: Have this now:

>- Ability to open a port for incoming internet traffic. If using a home wifi network, this may require changing the settings of your router.

docs/advanced/full-node.md
```diff
@@ -0,0 +1,63 @@
...
+- A safe place to leave your device running
+- If you need a silent node (hard disk drives can be loud), a SSD or USB drive
+
+## Guides
+
+Below are some guides from the community that walk you step-by-step through setting up a full node on various hardware and OSes. For generic high-level instructions, see the [Steps](#steps) section below.
```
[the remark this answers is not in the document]
s-ben: Done

docs/advanced/full-node.md
```diff
@@ -0,0 +1,63 @@
...
+
+- Access to the settings of your router (to allow outbound connections)
+- A safe place to leave your device running
+- If you need a silent node (hard disk drives can be loud), a SSD or USB drive
+
+## Guides
+
+Below are some guides from the community that walk you step-by-step through setting up a full node on various hardware and OSes. For generic high-level instructions, see the [Steps](#steps) section below. 
+
+- [Running a Decred Raspberry Pi Node](https://medium.com/@_Checkmatey_/running-a-decred-raspberry-pi-node-ac605b70c652): A more beginner friedly guide that uses a VNC server (remote desktop). This guide is intented to be a gentle introduction to using SSH on the command line.
```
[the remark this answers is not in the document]
s-ben: Done

docs/advanced/full-node.md
```diff
@@ -0,0 +1,63 @@
...
+- Access to the settings of your router (to allow outbound connections)
+- A safe place to leave your device running
+- If you need a silent node (hard disk drives can be loud), a SSD or USB drive
+
+## Guides
+
+Below are some guides from the community that walk you step-by-step through setting up a full node on various hardware and OSes. For generic high-level instructions, see the [Steps](#steps) section below. 
+
+- [Running a Decred Raspberry Pi Node](https://medium.com/@_Checkmatey_/running-a-decred-raspberry-pi-node-ac605b70c652): A more beginner friedly guide that uses a VNC server (remote desktop). This guide is intented to be a gentle introduction to using SSH on the command line. 
+- [My experience setting up a Decred full-node](https://medium.com/crypto-rocket-blog/my-experience-setting-up-a-decred-full-node-8a9bbf55bc30): Another beginner-friendly guide that walks through setting up a full node on a Raspberry Pi using an Ubuntu terminal on a Windows PC.
```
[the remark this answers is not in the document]
s-ben: Fixed

docs/advanced/full-node.md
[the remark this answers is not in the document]
s-ben: Have the below now. Though wondering, do we want to recommend inbound peers also, for the benefits they provide the network?

1. **Enable incoming connections**
	- `dcrd` will automatically begin downloading the blockchain and connecting to peers. However, if you want to help the network and have your node start validating transactions, you'll need to open a port for incoming internet traffic. This may require changing the settings of your router (to allow outbound connections). If this is the case:
		1. Find your [Local IP address](https://www.howtogeek.com/236838/how-to-find-any-devices-ip-address-mac-address-and-other-network-connection-details/) and [External IP address](https://whatismyipaddress.com/).
		1. Edit the settings of your router to open port 9108 (port forwarding). This [process](https://www.wikihow.com/Set-Up-Port-Forwarding-on-a-Router) depends on the type of router you have. 
	- If running a firewall, you'll also need to configure it to allow inbound connections on port 9108.

docs/advanced/full-node.md
s-ben: One day I will learn...

docs/advanced/full-node.md
s-ben: I assume you wanted this (it's hard to tell what line comments refer to):

>- If running a firewall, you'll also need to configure it to allow inbound connections on port 9108.

docs/advanced/full-node.md
s-ben: Good point. Added:

>1. **Leave running**
	-  Leave your node running, online, in a safe space, 24/7/365.

docs/advanced/full-node.md
s-ben: Was on the fence on that myself. Changed to 'Decred full node', as I think that is probably the convention actually.

docs/advanced/full-node.md
s-ben: Agreed. Removed

docs/advanced/full-node.md
s-ben: Removed bullet

docs/advanced/full-node.md
s-ben: While dcrd does keep a running tally, apparently there isn't a 'finished' message yet (per above PR). So keeping as it for now. We can update when dcrd 1.5 is realesed (along with a couple other minor changes).

s-ben:
@jholdstock @RichardRed0x have addressed your feedback. Looking pretty good now...Looks like barring additional feedback, the only outstanding question is on how to handle setting the external IP address.

docs/advanced/full-node.md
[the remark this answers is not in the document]
jholdstock: Looks good to me. We should remove the extra params (`&zoom=ikd7pc00-k0vn9c00&bin=day&axis=time`) from the url so the chart loads with the default zoom etc.

docs/advanced/full-node.md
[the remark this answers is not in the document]
jholdstock: Its still missing the `n`

docs/advanced/full-node.md
jholdstock: Yep thats good. I believe comments always appear directly under the line they are attached to.

docs/advanced/full-node.md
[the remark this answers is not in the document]
jholdstock: > However, if you want to help the network and have your node start validating transactions, you'll need to open a port for incoming internet traffic.

Opening the port is not required for validating transactions. I would suggest something like the following:

> To maximise your positive impact on the Decred network, it is important to allow inbound peers to connect to your node. Allowing inbound peers will allow new participants in the Decred network to connect to your node, and also enable your node to serve lightweight clients such as SPV wallets.

I found some nice reading on this topic - the first post from David in this forum thread
https://forum.decred.org/threads/how-dcrd-works-and-the-shortage-of-public-dcrd-nodes.1371/

Its a very good thread, lots in there which could be taken for this new article you are writing.

docs/advanced/full-node.md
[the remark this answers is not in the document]
s-ben: Done

docs/advanced/full-node.md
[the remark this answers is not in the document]
s-ben: Have gone with your suggested text. 

Wondering if it makes sense to add material from that [forum post](https://forum.decred.org/threads/how-dcrd-works-and-the-shortage-of-public-dcrd-nodes.1371/). We could maybe add finding your IP address or verifying inbound connection by looking for,
```[INF] BMGR: New valid peer 115.66.31.148:52400 (inbound) (/dcrwire:0.0.1/dcrd:0.0.8/)```

Too low level? Outdated?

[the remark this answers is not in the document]
Comment (jholdstock) [reply docs/advanced/full-node.md]: Good idea, we could keep it short and sweet just with something like:

> You will know when your node is correctly configured because you will see inbound peers in the logs. Inbound peers can also be checked by using the RPC `getpeerinfo`

File: docs/advanced/full-node.md (hunk 17 above)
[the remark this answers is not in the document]
Comment (s-ben) [reply docs/advanced/full-node.md]: Fixed!

File: docs/advanced/full-node.md (hunk 13 above)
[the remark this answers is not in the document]
Comment (s-ben) [reply docs/advanced/full-node.md]: Have added bullet:

>- If you have configured your node to allow inbound peers, you should see inbound peers in the logs. Inbound peers can also be checked using the Remote Procedure Call (RPC) command [getpeerinfo](../wallets/cli/dcrctl-rpc-commands.md).

**Comment by s-ben** [comment]
@jholdstock have addressed last outstanding comment (hopefully). I think we're good to go unless @davecgh thinks we should change the order of setting the external IP address ([comment](https://github.com/decred/dcrdocs/pull/989#discussion_r327534055)).

File: docs/advanced/full-node.md (hunk 6 above)
Comment (davecgh) [reply docs/advanced/full-node.md]: Yes, but it won't matter with 1.5.0 anymore for most cases anyway, because it will now automatically discover your external IP address based on what the other peers think you are.

That said, it can still be overridden with the `externalip` flag.

Comment (s-ben) [reply docs/advanced/full-node.md]: When will v1.5.0 be released, roughly? Unless it's going to happen soon, I'm inclined to just document it assuming v1.4.0. However, how else would one set the externalip? Only seeing how to pass externalip when restarting dcrd...would the user just restart it twice?

Comment (davecgh) [reply docs/advanced/full-node.md]: We're aiming for the RC of 1.5.0 within 2 weeks.  Correct you can only set externalip on the command line.

Comment (jholdstock) [reply docs/advanced/full-node.md]: OK lets not get caught up on this. We can just go with the text as-is, and we can revisit when 1.5 comes out.

File: docs/advanced/full-node.md [hunk 19]
```diff
@@ -0,0 +1,68 @@
...
+		1. Find your [Local IP address](https://www.howtogeek.com/236838/how-to-find-any-devices-ip-address-mac-address-and-other-network-connection-details/) and [External IP address](https://whatismyipaddress.com/).
+		1. Edit the settings of your router to open port 9108 (port forwarding). This [process](https://www.wikihow.com/Set-Up-Port-Forwarding-on-a-Router) depends on the type of router you have. 
+	- If running a firewall, you'll also need to configure it to allow inbound connections on port 9108.
+1. **Wait for dcrd to sync**
+	- Check a Decred block explorer such as [dcrdata](https://dcrdata.decred.org/) and wait until the block height displayed matches the block height shown locally.
+1. **Restart dcrd**
+	- Restart `dcrd` and pass it your external IP address using the following flag `--externalip=[your.external.ip]`
+	- If you don't wish to pass your IP address every time you restart `dcrd`, you can put it in dcrd's [configuration file](../wallets/cli/startup-basics.md#configuration-file-locations). \
```
jholdstock: Took one last pass over this and noticed a rogue `\` at the end of this line.

The PR is good to merge once this is remove.

s-ben: Removed

s-ben:
@jholdstock, removed spare `\`, should be good to go.

PR #991, opened 2019-09-24
multi: http -> https

dajohi:


.github/workflows/python.yml
```diff
@@ -17,7 +17,7 @@ jobs:
     - name: Install dependencies
       run: |
         python -m pip install --upgrade pip
-        pip install --user -r requirements.txt
```
jholdstock: `--user` was here to ensure packages are only installed for the current user. Removing `--user` may cause issues in future because packages may attempt to write into non-user locations, hence requiring the use of `sudo`.

Is there a good reason to remove `--user`?

PR #1003, opened 2019-10-14
New Glossary terms

jholdstock:
- Change
- Change Address
- Elliptic Curve
- Elliptic Curve Cryptography
- Smart Contract

### DIFF
--- a/docs/glossary.md
+++ b/docs/glossary.md
@@ -163,10 +163,10 @@ that may be essentially anywhere.
 Assuming N is sufficiently large, reversing this process is computationally
 infeasible.
 
-#### Elliptic Curve Cryptography
+#### Elliptic Curve Cryptography (ECC)
 
-Elliptic-curve cryptography (ECC) is an approach to public-key cryptography
-based on the algebraic structure of elliptic curves over finite fields.
+An approach to public-key cryptography based on the algebraic structure of
+elliptic curves over finite fields.
 
 ECC requires significantly smaller keys to achieve equivalent security compared
 to non-EC cryptography techniques - for example, a 256 bit ECC key is equivalent
@@ -383,8 +383,8 @@ A wallet mode in which only blocks related to addresses owned by the wallet are 
 
 #### Smart contract
 
-A smart contract is a self-executing contract with the terms of the agreement
-between buyer and seller being directly written into lines of code.
+A self-executing contract with the terms of the agreement between buyer and
+seller being directly written into lines of code.
 
 Once a smart contract has been written and broadcast to the Decred network, the
 code and the agreements contained therein exist across the distributed,
@@ -393,6 +393,9 @@ As a result, the code fully controls the execution of the contract, enabling
 trusted transactions and agreements to be carried out among disparate, anonymous
 parties without the need for a central authority, legal system, or external
 enforcement mechanism.
+
+The scripting language used to create Decred smart contracts is not [Turing
+complete](https://en.wikipedia.org/wiki/Turing_completeness).
 
 #### Stake transaction tree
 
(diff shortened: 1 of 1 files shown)


docs/glossary.md
```diff
@@ -50,6 +50,35 @@ In each block, five tickets are called to vote. In addition to votes on any open
 
 ---
 
+#### Change
+
+Decred has a concept of change, which is very similar to the concept of change
+when using physical cash.
+
+To explain by way of an example, imagine you have received 5 DCR in a single
```
s-ben: While this is a useful example, it doesn't fit well with the general form of the other definitions, which are more concise and don't have examples fleshed out. Also, I think this general example/explanation will available elsewhere on the net if someone is curious to learn more.

docs/glossary.md
```diff
@@ -322,6 +381,19 @@ A [simulation network](advanced/simnet.md) with very low difficulty, such that a
 
 A wallet mode in which only blocks related to addresses owned by the wallet are downloaded. When not in SPV mode, wallets must download all blocks in the blockchain to verify transactions (fully validating mode). SPV mode allows wallets to operate with less stringent hardware requirements and load significantly faster. SPV wallets cannot vote, but can purchase tickets and allocate voting rights to a [Voting Service Provider (VSP)](#voting-service-provider).
 
+#### Smart contract
+
+A smart contract is a self-executing contract with the terms of the agreement
```
s-ben: Convention in glossary is to start with an article. I.e. this,
>A smart contract is a self-executing contract with the terms of the agreement

Would be:
> A self-executing contract with the terms of the agreement

The other new entries need to be similarly changed to adhere to convention.

docs/glossary.md
```diff
@@ -322,6 +381,19 @@ A [simulation network](advanced/simnet.md) with very low difficulty, such that a
...
+between buyer and seller being directly written into lines of code.
+
+Once a smart contract has been written and broadcast to the Decred network, the
```
s-ben: I think we could shorten this to make it more concise like the other definitions. Also, I think we need to address the general confusion around the definition of smart contracts. Many people think smart contracts are only turning complete (or more so) contracts like Ethereum has. We don't want people thinking Decred is a "smart contract platform", when it only (for now) has smart contracts in the more limited definition shared with Bitcoin.

docs/glossary.md
```diff
@@ -118,6 +147,36 @@ A double-spend is an attack where the given set of coins are spent in more than
 
 ---
 
+#### Elliptic Curve
+
+In mathematics, an elliptic curve is a plane algebraic curve defined by an
```
s-ben: I think some of this definition (and the Elliptic Curve Cryptography definition) would be better placed in the Elliptic Curve Signature Algorithms [page](https://docs.decred.org/research/elliptic-curve-signature-algorithms/), which we can link to from here. This makes the glossary definitions more concise and in line with convention on this page, as well as makes the Elliptic Curve Signature page stronger.

jholdstock: The glossary entry is intended to be a high-level summary of the entire field of ECC. Target audience would be non-cryptographers/mathematicians who have seen the phrase ECC passed around, but have no idea what it means.

The ECSA page is diving into the specifics of one particular aspect of ECC - signature algorithms. Target audience for this page is a reader who already has a decent understanding of ECC and its intricacies.

Would suggest keeping these separate for that reason - unless you have any specific recommendations about what to move/update?

docs/glossary.md
jholdstock: Agreed

docs/glossary.md
jholdstock: Added a line

> The scripting language used to create Decred smart contracts is not [Turing complete](https://en.wikipedia.org/wiki/Turing_completeness).

PR #1008, opened 2019-10-22
Update logo to color

s-ben:
To visually distinguish the upcoming developer documentation from dcrdocs (they both use mkdocs and the same basic template), we have [changed the background color](https://github.com/jholdstock/dcrdevdocs/issues/40) for devdocs. However, we can further differentiate dcrdocs by updating the logo from monochrome (current) to full color. I think it looks better anyway. 

image

Thoughts?


jholdstock:
Looks good but please remove the old svg file which is no longer used.

s-ben:
@jholdstock logo removed.

PR #1011, opened 2019-10-22
Update the wallets/hardware-wallets.md page

teknico:
Changes:

- update the date of last update;
- split the table into one for Trezor and one for Ledger;
- add staking info;
- add rows about the Exodus wallet;
- add links and a References chapter.


teknico:
> The column for staking seems a bit redundant considering it only contains one value - "No".

The information that no hardware wallet currently supports staking seems relevant.

> Also the use of references here doesn't really fit the way refs are used in the rest of the docs.

The intention was to make it easier to check for status changes in the future, both for readers and maintainers.

> just including a single link out to exodus.io .

The reason for this is unclear.

Thanks for your consideration anyway.

PR #1018, opened 2019-10-30
Add forking terms to glossary

### LINKED ISSUE #1001: Add forking terms to Glossary
- Fork
- Hard Fork
- Soft Fork

s-ben:
Adds definitions for Fork, Hard Fork and Soft Fork to glossary ( Closes #1001 ).


docs/glossary.md
```diff
@@ -189,6 +189,14 @@ Tickets that reached the end of their window without being called to vote - thes
...
+
+It is also worth noting that chain forks can and do regularly occur without any code changes. This happens unintentionally when nodes in the network temporarily disagree on the state of the blockchain. Typically, these forks are short lived and resolve when the nodes reach consensus.
```
zubairzia0: It could be valuable to add a line regarding sustained forks as they can cause a great deal of damage and decred is well protected from them.

s-ben: Added a line that defines a sustained chain split. Definition taken from @noahpierau's article that explains Decred's advantages here (which I also link to):

>It is also worth noting that chain forks can and do regularly occur without any code changes. This happens unintentionally when nodes in the network temporarily disagree on the state of the blockchain. Typically, these forks are short lived and resolve when the nodes reach consensus. In [sustained chain splits](https://blog.goodaudience.com/blockchain-forks-and-chain-splits-why-we-should-avoid-them-f54c693a90f1) such as Bitcoin Cash, there is sufficient economic and hash power to support multiple competing versions of the blockchain for an extended period of time.

docs/glossary.md
```diff
@@ -189,6 +189,14 @@ Tickets that reached the end of their window without being called to vote - thes
 
 A mechanism for obtaining free (testnet) coins.
 
+#### Fork
+
+A programming term for an open-source code modification. The modified code is considered a “fork” of the original code. Typically, forks contain modifications that significantly alter the original code. For example, the Bitcoin Cash fork of the Bitcoin blockchain raised the block size from 1Mb to 8 Mb. If the state of the blockchain is also forked, as was the case with Bitcoin Cash, it is also considered a “chain fork”. In this case, owners of coins on the original chain will have equivalent coins on the new chain when the chains diverge. Decred began its life as a source code fork of [btcd](https://github.com/btcsuite/btcd), an alternate full node implementation of Bitcoin. Significant changes were then made before launch, including modifications to the consensus rules, a Proof-of-Stake (PoS) layer to enable on-chain governance, a new hashing algorithm and other innovations. It is arguable whether Decred is a source code fork, considering the scope of modifications made. Decred is not a chain fork, as it did not copy the state of the Bitcoin blockchain, and instead started from its own genesis block.
```
davecgh: I see both `1Mb` and ` 8 Mb`, one with a space and one without.  They should be consistent, imo.

docs/glossary.md
```diff
@@ -397,6 +411,10 @@ enforcement mechanism.
 The scripting language used to create Decred smart contracts is not [Turing
 complete](https://en.wikipedia.org/wiki/Turing_completeness).
 
+#### Soft Fork
+
+A change to consensus rules that is backwards compatible. Nodes running older versions without the new rules can still validate transactions and blocks created with the new rules. New transaction types, for instance, can often be added with soft-forking changes. However, only nodes using the new rules will be able to understand the new transaction type.
```
davecgh: I think really should properly call out a couple of important features:

A change to consensus rules that is backwards compatible.  **The backwards compatibility is achieved by tricking old nodes into believing they are faithfully and fully validating the transactions and blocks even though they are not.**  Nodes running older versions without the new rules can still **partially** validate transactions and blocks created with the new rules.  New transaction types, for instance, can often be added with soft-forking changes. However, only nodes using the new rules will be able to understand the new transaction type.

s-ben: Good point. Changed to:

>A change to consensus rules that is backwards compatible. Nodes running older versions without the new rules can still validate transactions and blocks created with the new rules. However, nodes running older versions can only partially validate transactions and blocks created using the new rules. For instance, if a new transaction type is added in a soft-fork, only nodes using the new rules will be able to understand the new transaction type.

docs/glossary.md
s-ben: Good catch. Went with no space (`8Mb`).

docs/glossary.md
davecgh: I really think it should call out the reality that is tricking old nodes into believing they are faithfully and fully validating the transactions and blocks even though they are not.  That is a super important distinction that I don't see being made here.

s-ben: I hesitate to add the words "tricking" and "faithfully" here, because while I agree with these characterizations, they are characterizations, not objective facts (what technical docs should generally stick to). Tricking implies deception, and code cannot be deceived, only the humans running it. And we can't strictly assume the humans running old code don't know what the deal is. My knowledge is limited here though, and open to counter argument. I just think we should make these arguments as objectively and dispassionately as possible. E.g. if there's another example, or negative consequence that a reader would reasonably interpret as "tricking", that would be good to add.

davecgh: I see your point, but it is an objective fact that old nodes believe they are fully validating the transactions and blocks even though they are not just in the same way it is an objective fact that an old node will reject blocks that are using rules that are the result of a hard fork.

I suppose you can leave out "tricking" and "faithfully" if you feel they're more in the characterization realm, although, I would still argue they are objectively true.  I would be happy to just specifically call out "Under a soft fork, old nodes believe they are fully validating the transactions and blocks even though they are not."

s-ben: Well, I could argue that software doesn't have beliefs systems either:)

However, is has been common since the invention of programming for coders to anthropomorphize their code, imagining code objects 'believing' and 'looking', etc. So I think this falls squarely into standard, objective terminology. How does this sound?

>A change to consensus rules that is backwards compatible. Nodes running older versions without the new rules can still validate transactions and blocks. However, nodes running older versions can only partially validate transactions and blocks created by nodes using the new rules. This is because old nodes believe they are fully validating transactions, however, they are not able to undstand transactions created using the new rules. For instance, if a new transaction type is added via soft fork, only nodes using the new rules will be able to understand the new transaction type.

davecgh: Sounds good.  I thought you had already made the change and approved the PR, but I'll do it again once you make it.

s-ben: Oops. That is my normal process. Committed and pushed new text.

docs/glossary.md
jholdstock: I think this is still incorrect - should be mega**bytes**, not mega**bits** - MB, not Mb

docs/glossary.md
```diff
@@ -189,6 +189,14 @@ Tickets that reached the end of their window without being called to vote - thes
 
 A mechanism for obtaining free (testnet) coins.
 
+#### Fork
```
jholdstock: As this section is describing **software** forks and not **chain** forks, using Bitcoin/Bitcoin Cash as an example is rather confusing because that represents both a software and a chain fork. For ultimate clarity, I would suggest using an example completely outside of cryptocurrency - eg. Ubuntu as a fork Debian, LibreOffice as a fork of OpenOffice, Plex as a fork of XBMC, or if you want to go really retro and shout-out to some of the c0 guys, OpenBSD as a fork of NetBSD.

docs/glossary.md
```diff
@@ -189,6 +189,14 @@ Tickets that reached the end of their window without being called to vote - thes
...
+
+It is also worth noting that chain forks can and do regularly occur without any code changes. This happens unintentionally when nodes in the network temporarily disagree on the state of the blockchain. Typically, these forks are short lived and resolve when the nodes reach consensus. In [sustained chain splits](https://blog.goodaudience.com/blockchain-forks-and-chain-splits-why-we-should-avoid-them-f54c693a90f1) such as Bitcoin Cash, there is sufficient economic and hash power to support multiple competing versions of the blockchain for an extended period of time.
+
+Source code forks can be [hard forks](#hard-fork) or [soft forks](#soft-fork). Soft forks are backwards compatible, whereas hard forks are not.
```
jholdstock: Source code forks are not hard/soft. This is referring to chain forks.

docs/glossary.md
```diff
@@ -189,6 +189,14 @@ Tickets that reached the end of their window without being called to vote - thes
 
 A mechanism for obtaining free (testnet) coins.
 
+#### Fork
+
+A programming term for an open-source code modification. The modified code is considered a “fork” of the original code. Typically, forks contain modifications that significantly alter the original code. For example, the Bitcoin Cash fork of the Bitcoin blockchain raised the block size from 1Mb to 8Mb. If the state of the blockchain is also forked, as was the case with Bitcoin Cash, it is also considered a “chain fork”. In this case, owners of coins on the original chain will have equivalent coins on the new chain when the chains diverge. Decred began its life as a source code fork of [btcd](https://github.com/btcsuite/btcd), an alternate full node implementation of Bitcoin. Significant changes were then made before launch, including modifications to the consensus rules, a Proof-of-Stake (PoS) layer to enable on-chain governance, a new hashing algorithm and other innovations. It is arguable whether Decred is a source code fork, considering the scope of modifications made. Decred is not a chain fork, as it did not copy the state of the Bitcoin blockchain, and instead started from its own genesis block.
```
jholdstock: Decred code is a fork of btcsuite. I do not think this is arguable, I think this is a fact. Fork is a well defined term - using a copy of the source code from one project as the basis of another project - and that is exactly what happened with btcsuide and Decred.

docs/glossary.md
jholdstock: The section actually blends the two definitions, making it quite hard to distinguish which is which. 

Perhaps we could simplify by splitting these very clearly into two distinct sections:

> The term fork has two common meanings:
>  - Source code forks are...
>  - Chain forks are ....

docs/glossary.md
jholdstock: Curious for your opinion on this @davecgh?

davecgh: It would probably be most accurate to say that Decred began its life as a source code fork of [btcsuite](https://github.com/btcsuite), because more than just dcrd/btcd required modifications, but, the core network is indeed dcrd, so I didn't bother nitpicking that on my review.

In terms of the bit in question here about it being arguable, I agree from a purely technical standpoint, but I also understand what the text is conveying, so it didn't really stand out as something major to me.   The issue is that typically when you talk about something being a fork of something else, it gives the impression that the are very few differences between the two, which is really not the case here at all.  The two codebases have _significantly_ diverged at this point that still referring to it as such doesn't really properly describe the situation.

Perhaps something like "...were then made before launch, including ... a new hashing algorithm and other innovations.  Additionally, since launch, the code base has continued to diverge from btcsuite so significantly that it is a standalone code base."

docs/glossary.md
s-ben: I see what you mean. Took a stab at rewording the first paragraph:

>In the context of blockchains, the term fork has two common meanings; it can refer to either a code fork or a chain fork. A [code fork](https://en.wikipedia.org/wiki/Fork_(software_development)) is when developers take a copy (fork) of the source code of a project and begin independently developing it, creating a separate version of the software. For instance, the OpenBSD operating system is a fork of NetBSD. If the state of the blockchain is also forked, as was the case with Bitcoin Cash, it is also considered a “chain fork”. In this case, owners of coins on the original chain will have equivalent coins on the new chain when the chains diverge. Decred began its life as a source code fork of [btcd](https://github.com/btcsuite/btcd), an alternate full node implementation of Bitcoin. Significant changes were then made before launch, including modifications to the consensus rules, a Proof-of-Stake (PoS) layer to enable on-chain governance, a new hashing algorithm and other innovations. It is arguable whether Decred is a source code fork, considering the scope of modifications made. Decred is not a chain fork, as it did not copy the state of the Bitcoin blockchain, and instead started from its own genesis block.

docs/glossary.md
s-ben: Upon rereading, I see @jholdstock's point. I think the fact that we say "Decred **began its life** as a source code fork of [btcd](https://github.com/btcsuite/btcd)," and then describe the significant changes, is enough to make the argument that it is its own standalone code base now. We could keep wordsmithing this, but this entry is getting long already, and I think it's better to be concise. Here's what I have now:

>Decred began its life as a source code fork of [btcd](https://github.com/btcsuite/btcd), an alternate full node implementation of Bitcoin. Significant changes were then made before launch, including modifications to the consensus rules, a Proof-of-Stake (PoS) layer to enable on-chain governance, a new hashing algorithm and other innovations. Decred is not a chain fork, as it did not copy the state of the Bitcoin blockchain, and instead started from its own genesis block.

docs/glossary.md
s-ben: Good point. Have just gone with 'fork':

>Forks can be [hard forks](#hard-fork) or [soft forks](#soft-fork). Soft forks are backwards compatible, whereas hard forks are not.

s-ben:
@jholdstock, made a couple tweaks based on your suggestions. Lgtm

PR #1019, opened 2019-11-04
Adds Coin type and HD Wallet to Glossary

### LINKED ISSUE #922: Coin type
```
[WRN] WLLT: Wallet contains addresses derived for the legacy BIP0044 coin type and seed restores may not work with some other wallet software
```

Since this leaks onto the user it is no longer an internal dev thing.

Please tell me what I need to know (as a user) about this coin type thing. What is it, what are 20 and 42, what is BIP44 and SLIP44, why did it change, and most importantly, what do I need to do if I see this warning, and what happens if I don't.

Add Coin type to glossary.

imestin:
Closes #922


imestin:
Problem with automatic tests

PR #1020, opened 2019-11-04
Change/signature script isolation

### LINKED ISSUE #999: Update Signature Script Isolation and Fraud Proofs page
In the Signature Script Isolation and Fraud Proofs [page](https://docs.decred.org/research/signature-script-isolation-and-fraud-proofs/), we say:

>It is now being proposed for Bitcoin as a soft fork referred to as “Segregated Witness”

This should be updated to say that the soft fork happened. But also wondering if there is other outdated info on this page?

This line also might need updating after [DCP-0005](https://github.com/davecgh/dcps/blob/dcp0005/dcp-0005/dcp-0005.mediawiki#) passes (assuming it does):

>In addition, fraud proofs, as suggested for Bitcoin’s soft fork3, are set by miners and also committed to as part of the data in the merkle tree.

imestin:
Closes #999


imestin:
Merge conflict

PR #1021, opened 2019-11-04
Updates part about Segregated Witness and introduces DCP-0005

### LINKED ISSUE #999: Update Signature Script Isolation and Fraud Proofs page

imestin:
Closes #999

### DIFF
--- a/docs/research/signature-script-isolation-and-fraud-proofs.md
+++ b/docs/research/signature-script-isolation-and-fraud-proofs.md
@@ -2,7 +2,7 @@
 
 ---
 
-To prevent transaction malleability, the ability to generate a transaction with the same input references and outputs and yet a different transaction ID, input scripts have been removed from the calculation of the transaction hash. The origins of this modification have been controversial, although it appears to have been implemented in both CryptoNote coins and sidechains in the past[^1] [^2]. It is now being proposed for Bitcoin as a soft fork referred to as "Segregated Witness"[^3]. As in the Elements sidechains implementation, commitments to the witness data are included in the Merkle tree of the block[^2]. In addition, fraud proofs, as suggested for Bitcoin's soft fork[^3], are set by miners and also committed to as part of the data in the Merkle tree.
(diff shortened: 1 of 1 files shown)


docs/research/signature-script-isolation-and-fraud-proofs.md
```diff
@@ -2,7 +2,9 @@
...
+To prevent transaction malleability, the ability to generate a transaction with the same input references and outputs and yet a different transaction ID, input scripts have been removed from the calculation of the transaction hash. The origins of this modification have been controversial, although it appears to have been implemented in both CryptoNote coins and sidechains in the past[^1] [^2]. After the activation of the “Segregated Witness” [^3] soft fork, it is implemented in Bitcoin as well. As in the Elements sidechains implementation, commitments to the witness data are included in the merkle tree of the block[^2]. In addition, fraud proofs, as suggested for Bitcoin's soft fork[^3], are set by miners and also committed to as part of the data in the merkle tree.
```
s-ben: is is implemented in -> it was implemented in

s-ben: capitalize 'Merkle'

docs/research/signature-script-isolation-and-fraud-proofs.md
```diff
@@ -2,7 +2,9 @@
...
+
+If DCP-0005 [^4] activates, Decred will include fraud proofs in the block header. The name of the new field will be CommitmentRoot. Each leaf of the Merkle tree is intended to be a commitment to some arbitrary data which will typically be some structure that employs techniques that also allow compact fraud and inclusion proofs to be constructed such as Fenwick trees, Golomb-Rice filters, and additional Merkle trees. The number of short proofs will be extensible.
```
s-ben: This will be good to add once DCP-0005 passes (presumably it will, but we can't be sure). We never want to use the future tense in technical documentation. E.g. we don't want to 'if X' or 'when Y'. This runs the danger of either X not happening, or it happens and the docs don't get updated (more common). Let's take this out for now. Put it back in once DCP-0005 passes.

imestin: I removed it.

imestin:
Problem with automatic tests

PR #1023, opened 2019-11-17
Adds Coin type and HD Wallet to Glossary

### LINKED ISSUE #922: Coin type

imestin:
Closes #922


docs/glossary.md
```diff
@@ -221,6 +227,10 @@ A cryptographic function that produces a fixed-size hash value from variable-siz
 
 The number of hashes per second computed by miners on the network.
 
+#### HD Wallet
```
s-ben: Need to spell out acronym in header, i.e. 'Hierarchical Deterministic (HD) Wallet'

docs/glossary.md
```diff
@@ -221,6 +227,10 @@ A cryptographic function that produces a fixed-size hash value from variable-siz
...
+
+A wallet that uses the Hierarchical Deterministic (HD) protocol, that was defined in [BIP-0032](https://github.com/bitcoin/bips/blob/master/bip-0032.mediawiki). In HD wallets, a special parent seed is used to deterministically create child private keys for addresses. This means that the user typically only needs to back up the seed phrase.
```
s-ben: that was defined in -> which was defined in...

docs/glossary.md
```diff
@@ -87,6 +87,12 @@ Outputs from coinbase and stakebase transactions cannot be spent until the coinb
 
 Every block which is mined contains a single coinbase transaction. This transaction will only have one input, and that is newly created Decred which constitutes the PoW and Treasury portions of the block reward. Coinbase transactions are part of the regular transaction tree, which means that they will be rejected if proof-of-stake voters vote to reject the block which contains them. Decred created in coinbase transactions cannot be spent until the coinbase maturity period has passed.
 
+#### Coin type
+
+A constant, set for each cryptocoin for the BIP32 path. BIP-0032 introduces [HD Wallets](#hd-wallet). In the BIP32 path, every coin has an index. The index of Decred is 42. Registered coins are stored in a Satoshi Labs Improvement Proposal, in [SLIP-0044](https://github.com/satoshilabs/slips/blob/master/slip-0044.md).
```
s-ben: We say 'the BIP32 path', but what does path refer to here?

s-ben: >are stored in a Satoshi Labs Improvement Proposal, in [SLIP-0044](https://github.com/satoshilabs/slips/blob/master/slip-0044.md).

This wording is redundant. Suggest something like "..are stored in [SLIP-0044](https://github.com/satoshilabs/slips/blob/master/slip-0044.md)."

It would be nice to spell out the Satoshi Labs Improvement Proposal (SLIP) acronym here, but can't think of a non awkward way to do it. I think it's OK not to in this case. If the reader clicks on the link they'll see it spelled out. Alternately, we can use the current wording, but remove the 'in'. So, "in a Satoshi Labs Improvement Proposal, [SLIP-0044](https://github.com/satoshilabs/slips/blob/master/slip-0044.md)."

docs/glossary.md
```diff
@@ -87,6 +87,12 @@ Outputs from coinbase and stakebase transactions cannot be spent until the coinb
 
 Every block which is mined contains a single coinbase transaction. This transaction will only have one input, and that is newly created Decred which constitutes the PoW and Treasury portions of the block reward. Coinbase transactions are part of the regular transaction tree, which means that they will be rejected if proof-of-stake voters vote to reject the block which contains them. Decred created in coinbase transactions cannot be spent until the coinbase maturity period has passed.
 
+#### Coin type
+
+A constant, set for each cryptocoin for the BIP32 path. BIP-0032 introduces [HD Wallets](#hd-wallet). In the BIP32 path, every coin has an index. The index of Decred is 42. Registered coins are stored in a Satoshi Labs Improvement Proposal, in [SLIP-0044](https://github.com/satoshilabs/slips/blob/master/slip-0044.md). 
+
+Before SLIP44, the list of cryptocurrencies did not exist, the only 2 types were Bitcoin and Bitcoin Testnet. A wallet which contains addresses that were generated by the legacy BIP44 method, would contain addresses that the wallet software couldn’t see.
```
s-ben: "the only 2" -> "the only two"

Typically, when a number is in the single digits (<10), it's spelled out.

s-ben: >method, would contain

Remove comma. While I agree a pause kind of makes sense here, it's not technically grammatically correct, and the sentence reads OK without the comma.

docs/glossary.md
imestin: My problem is that there are too many redirections, my thought was that user will research BIP-0032 if he/she is interested. (Link is under HD Wallet)

It's true that it would be good to see a BIP32 path in the beginning of this article, I will try to do something about this.

jholdstock:
Superseded by #1057

PR #1027, opened 2019-11-25
Polish Privacy Guide

jholdstock:

PR #1029, opened 2019-11-26
Finishing touches on Privacy guide

jholdstock:
Implementing both of the suggestions from @s-ben in #1027 

The rest of the changes are just whitespace/line length

### DIFF
--- a/docs/privacy/general-privacy.md
+++ b/docs/privacy/general-privacy.md
@@ -32,7 +32,13 @@ your personal activity.
 An observer may be able to deduce other transactions which you have participated
 in, or they may be able to deduce your total balance.
 
-One possible way your personal identity could be linked to your blockchain activity is through the use of a block explorer. The administrator of the block explorer may be able to view your IP address and which addresses or transactions you are browsing, and they may infer that you own those addresses or transactions. A possible solution for this is to use blockchain explorers exclusively over Tor.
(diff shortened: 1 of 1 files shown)


xaur:
@jholdstock why wrap the lines manually?

jholdstock:
Many reasons tbh

- horizontal scrolling is the work of the devil
- shorter lines are significantly easier to update/review/provide feedback upon
- GitHub does not render diffs on very long lines
- git itself calculates diffs on a line-by-line basis, so splitting lines actually makes the repo smaller/faster to work with
- splitting lines enables the use of `git add --patch`

xaur:
Thanks, that is some food for thought @jholdstock!

Personally I prefer no manual line wrapping in text (in code it is a must ofc), so 1 line = 1 paragraph. The main issue with manual line wrapping is that it has issues with automatic reflowing of text (because now line breaks are hardcoded).

Imo any decent text editor should be able to wrap lines. All editors I used had this feature and I never had to scroll horizontally. I use Git with CLI + Meld for comparing files and reviewing changes and never had issues with reviewing either. GitHub's differ doesn't always diff long lines nicely, although Rich Diff mode seems to handle it (at least for Markdown).

Git repo size and `git add --patch` are interesting points I didn't consider. Do you have any numbers on possible size savings?

Issue #1041, opened 2019-12-31
Replace forum links


degeri:
All the links that point to the forum should be removed or updated. 

https://github.com/decred/dcrdocs/search?q=https%3A%2F%2Fforum&unscoped_q=https%3A%2F%2Fforum

jholdstock:
I havent removed all of the forum links in #1044 because most of them are within FAQ pages, and I think the FAQ pages should be updated as part of a larger change. Many of the FAQs are overly verbose, contain information which is duplicated elsewhere, or are completely outdated due to new developments.

Its worth reviewing the FAQ pages one by one and carefully updating, rather than just cutting out the links without any extra thought

s-ben:
Wondering if maybe we could make a goof-first-issue for the twitter feed? Might be good because it's an excuse to learn more about the project. Might be bad because it could lead to a lot of back and forth?
https://twitter.com/dcrgoodfirst

degeri:
We should avoid GFI that require alot of coordination

PR #1047, opened 2020-01-09
Remove forum links from wallet/seed FAQ

jholdstock:
Part of #1041 

I removed this question
```
How do I import a key that is in wallet import format (WIF)?
```
It was only a recurrent issue during the airdrop and is not so relevant today

### DIFF
--- a/docs/faq/wallets-and-seeds.md
+++ b/docs/faq/wallets-and-seeds.md
@@ -12,21 +12,40 @@ The safest method of storing your seed is on a physical medium, eg. written on p
 
 #### 2. How can I convert my wallet seed hex to seed words?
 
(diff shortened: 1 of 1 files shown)


s-ben:
Looks good. My one quibble is this run-on sentence:

>To put this in perspective, assuming there are 7 billion people on the planet and each person owned 10 computers and each one of those computers could test a billion possibilities a second and that you could find the solution on average after testing only 50% of the total possibilities, it would still take 26x10^48 (that’s 26 trillion trillion trillion trillion) years to brute-force a single seed.

Suggested rewording:

>To put this in perspective, assuming there are 7 billion people on the planet, with each person owning 10 computers, and each computer testing a billion possibilities a second, it would still take approximately 26x10^48 (that’s 26 trillion trillion trillion trillion) years to brute-force a single seed.

jholdstock:
Yeah thats a davec classic, copied from the forum because we are losing the link to the original source. Will update

PR #1049, opened 2020-01-09
Update PoW intro. Mention fees.

### LINKED ISSUE #1048: PoW Overview section does not mention transaction fees or ticket fees
The PoW section does not mention the transaction fees or the ticket fees that are awarded to miners.
https://docs.decred.org/mining/overview/

jholdstock:
Closes #1048

### DIFF
--- a/docs/mining/overview.md
+++ b/docs/mining/overview.md
@@ -6,31 +6,25 @@
 
 Proof-of-work mining, more commonly referred to as PoW mining, is the activity
 of committing your computer’s hardware and resources to process network
-transactions and build the blocks that make up the blockchain in the Decred
-network. Each time a block is created (by a miner), which is then approved by 
-ticket holders, the miner receives a block reward (newly created DCR). If the
-block is rejected by ticket holders, the miner does not receive a reward. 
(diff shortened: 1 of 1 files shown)


s-ben:
New text looks good. Only thing I'm wondering is why we're removing the chart showing the breakdown of the block rewards between PoW/PoS/Treasury. I know we have it elsewhere, but perhaps someone might find that info useful here? 

Also, the `- for example` seems a little wonky?

>This power may be used if the miner has constructed a block contrary to the best interests of the network - for example, an empty block when there are transactions waiting to be processed in the mempool.

I'm a grammar rebel, so I'm OK as long as it reads well (which I think it does). But the more compliant version would probably be:

>This power may be used if the miner has constructed a block contrary to the best interests of the network. For example, mining an empty block when there are transactions waiting to be processed in the mempool.

jholdstock:
I've removed the hyphen as suggested.

For me, the table didn't really seem to fit here, it seems too detailed and specific for an introductory paragraph. As you mentioned, that information is already present in many places throughout the docs. Linking to the inflation page which contains full details of how the block reward is split, and the justification, seems to be a better solution.

PR #1053, opened 2020-01-14
Refresh "How to Vote" page

### LINKED ISSUE #1050: Update How to Vote page to make it generic
In the [How to Vote page](https://docs.decred.org/governance/consensus-rule-voting/how-to-vote/) detailing voting on consensus changes, we reference "vote version 5". However, this will become out of date with subsequent votes. We should make the text generic so we don't have to update it after every vote (which we'll invariably forget to do).

### LINKED ISSUE #1051: Fix broken image How to Vote page
In the [How to Vote page](https://docs.decred.org/governance/consensus-rule-voting/how-to-vote/), an image is 404'ing. Fix image and, if needed, change the text introducing it (it may be out of date?).

jholdstock:
- Update VSP voting preferences image to new VSP UI design
- Re-add Decrediton voting preferences image (Closes #1051)
- Rename both images so they have standardised names
- Use proper markdown image tags rather than raw html
- Remove version specific examples from Solo Voting section (Closes #1050)


s-ben:
Looking good. Just have a couple suggestions for polishing old text.

>The getvotechoices command returns JSON describing the agendas..

'Returns JSON' is vague. Suggest:

>The getvotechoices command returns a JSON payload describing the agendas..

Suggested rewording for clarity on the last sentence:

>`"agendaid"` is found via the `getvotechoices` command above, and `"choiceid"`
can be `yes`, `no`, or `abstain`.

to:

>You can get the `agendaid` for votes using the above `getvotechoices` command. The `choiceid` field can be set to `yes`, `no`, or `abstain`.

jholdstock:
Thanks, I've implemented both suggestions

PR #1075, opened 2020-02-28
Cleanup dcrctl rpc dcrwallet commands

dezryth:
Removes duplicate walletlock command from list of RCP commands


s-ben:
@dezryth, looks good, but it appears there are two separate changes here: `Added social links and modified question in FAQ regarding storing see… ` and `removes duplicate walletlock command entry`. If this is intentional, can you update the PR title and description? If not, can you break into two PRs?

dezryth:
> @dezryth, looks good, but it appears there are two separate changes here: `Added social links and modified question in FAQ regarding storing see… ` and `removes duplicate walletlock command entry`. If this is intentional, can you update the PR title and description? If not, can you break into two PRs?

Thanks s-ben. I'm a bit of a git noob so I will see what I can do; not sure what happened here. I'll try to open a cleaner pull request.

PR #1079, opened 2020-03-11
Rename Inflation page

### LINKED ISSUE #1074: Rename Inflation page
Term ["inflation"](https://en.wikipedia.org/wiki/Inflation) is ambiguous and often refers to "prices inflation". Our [Inflation page](https://docs.decred.org/advanced/inflation/) has nothing to do with prices but instead is talking about "inflation of the monetary supply".

There are more specific terms for the page: Emission, Emission schedule, [Monetary inflation](https://en.wikipedia.org/wiki/Monetary_inflation), Issuance (schedule), Creation of (money|currency), [Money supply](https://en.wikipedia.org/wiki/Money_supply), etc. I would pick Emission.

jholdstock:
Open to debate if anybody disagrees with the new name "Issuance"

Closes #1074


jzbz:
Yup, more accurate, go for it.

PR #1091, opened 2020-04-27
Update dep versions

jholdstock:


jholdstock:
I think its a typical thing when using `/usr/local` for deps. All python projects on the same machine will use deps from the same directory which causes conflicts. I'd recommend working in [virtual environments](https://docs.python.org/3/tutorial/venv.html) to avoid this

PR #1106, opened 2020-06-18
Add page about LN backups

matheusd:
Ideally the page should show Decrediton screenshots but the relevant PR is still open (https://github.com/decred/decrediton/pull/2535) and the UI design isn't final yet so after this is merged I'll open an issue to track and add the screenshots that reflect an actual released version.

### DIFF
--- a/docs/lightning-network/backups.md
+++ b/docs/lightning-network/backups.md
@@ -1,0 +1,125 @@
+# LN Data Backup
+
+---
+
+!!! danger "Critical Information to Prevent Fund Loss"
+
+    In order to redeem funds from a channel in case of seed restores, LN users need to maintain an up to date backup of data by way of an SCB file.
+
+!!! info "SCB Decryption Errors"
+
+    An SCB file is only valid for the exact same wallet/account that generated it. Trying to restore it on a different wallet or account will cause a generic decryption error.
+
+!!! info "Watchtowers Required"
+
+    For complete safety from possible malicious remote nodes, users should also use third party, reputable watchtowers to ensure that if the remote nodes executes the DLP protocol with an invalid state they are correctly punished for it.
+
+
+Due to its nature as a second-layer network, Lightning Network related wallet data is **not** stored in the blockchain itself. This means that the standard wallet seed is **not sufficient** to restore the LN balance of a wallet in case of a re-seed.
+
+LN users need to **also** regularly and safely store the _Static Channel Backup_ (SCB) file so that a restored lightning wallet might be used to close the channels opened by the previous wallet.
+
+The backup should be updated every time a channel is opened or closed on the local LN node (including when a _remote_ node opens a channel back to the local node).
+
+The rest of this document details what exactly the SCB-based backup method entails, how to perform backups and restores using it and its limitations.
+
+## Overview of Backup Issues for LN nodes
+
+The next sections provide a high level overview of _why_ backups are needed for Lightning Network nodes and how they work.
+
+For the operational instructions on generating and applying backups see the section on [Operating SCB Backups](#operating-scb-backups).
+
+### Preamble: What data does an LN wallet store?
+
+This can be skipped for those familiar with how LN operates. 
+
+Payments over the Lightning Network happen across _channels_. An LN channel can be understood as an agreement between two entities (nodes) to allow payments up to some pre-defined amount.
+
+While a lightning channel is realized as an on-chain output, spendable by a multisig 2-of-2 signature script, the channel state evolves completely off-chain by having _relative balances_ and _in-flight payments_ (HTLCs) only known to the parties of channel.
+
+This means that while other participants of the network know that a given channel exists (due to the output being stored and confirmed on the blockchain), the data about which portion (how many coins) each of the two nodes own in the channel as well as which payments are currently outstanding is only known to them.
+
+Every sequential update to the channel state is summarized by a corresponding _commitment transaction_: a transaction that is usually kept off-chain (never broadcasted) but that works as a sort of insurance that either party can recover their funds if the other node stops working properly.
+
+This transaction is updated and re-signed every time a new payment changes the relative balances of the channel. It is then stored on a local database, along with some other metadata such as the current state number, the specific keys used to sign this update and information about older, revoked states.
+
+A lightning node must **never** publish a state older than the most recent one agreed by the two parties of a channel. Due to how the LN protocol works, publishing the commitment transaction of an old, revoked state means that party forfeits their funds to the other node, who can then send the _entire_ channel amount to a regular key under their sole control.
+
+Notice that since the commitment data is never published to the blockchain during the regular operation of an LN node, any event that causes loss of data means the node cannot recover their funds if the counterparty disappears.
+
+It also means the node cannot even advance the state of the channel, since it doesn't know what the current state _is_ to begin with. 
+
+### Unsafety of copying the dcrlnd DB
+
+!!! danger "Do not perform full-db backups"
+
+    This section details why full-db backups are **not safe** in the current implementation of `dcrlnd`. 
+
+    Do **not** perform full-db backups unless you have very deep knowledge on how LN operates and how `dcrlnd` is implemented and the _complete_ loss of funds that might occurr when using an incorrect db backup.
+
+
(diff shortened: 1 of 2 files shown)


docs/lightning-network/backups.md
```diff
@@ -0,0 +1,125 @@
...
+
+The rest of this document details what exactly the SCB-based backup method entails, how to perform backups and restores using it and its limitations.
+
+## Overview of Backup Issues for LN nodes
+
+The next sections provide a high level overview of _why_ backups are needed for Lightning Network nodes and how they work.
+
+For the operational instructions on generating and applying backups see the section on [Operating SCB Backups](#operating-scb-backups).
+
+### Preamble: What data does an LN wallet store?
```
jholdstock, line 32: The prefix "Preamble:" seems strange here. I think we can remove it

docs/lightning-network/backups.md
```diff
@@ -0,0 +1,125 @@
...
+
+This can be skipped for those familiar with how LN operates. 
+
+Payments over the Lightning Network happen across _channels_. An LN channel can be understood as an agreement between two entities (nodes) to allow payments up to some pre-defined amount.
+
+While a lightning channel is realized as an on-chain output, spendable by a multisig 2-of-2 signature script, the channel state evolves completely off-chain by having _relative balances_ and _in-flight payments_ (HTLCs) only known to the parties of channel.
+
+This means that while other participants of the network know that a given channel exists (due to the output being stored and confirmed on the blockchain), the data about which portion (how many coins) each of the two nodes own in the channel as well as which payments are currently outstanding is only known to them.
```
jholdstock, line 40:
```suggestion
This means that while other participants of the network know that a given channel exists (due to the output being stored and confirmed on the blockchain), the data about which portion (how many coins) each of the two nodes own in the channel, as well as which payments are currently outstanding, is only known to them.
```

docs/lightning-network/backups.md
```diff
@@ -0,0 +1,125 @@
...
+
+Every sequential update to the channel state is summarized by a corresponding _commitment transaction_: a transaction that is usually kept off-chain (never broadcasted) but that works as a sort of insurance that either party can recover their funds if the other node stops working properly.
```
jholdstock, line 42: never b
```suggestion
Every sequential update to the channel state is summarized by a corresponding _commitment transaction_: a transaction that is usually kept off-chain (never broadcast) but works as a sort of insurance that either party can recover their funds if the other node stops working properly.
```

docs/lightning-network/backups.md
```diff
@@ -0,0 +1,125 @@
...
+
+It also means the node cannot even advance the state of the channel, since it doesn't know what the current state _is_ to begin with. 
+
+### Unsafety of copying the dcrlnd DB
+
+!!! danger "Do not perform full-db backups"
+
+    This section details why full-db backups are **not safe** in the current implementation of `dcrlnd`. 
+
+    Do **not** perform full-db backups unless you have very deep knowledge on how LN operates and how `dcrlnd` is implemented and the _complete_ loss of funds that might occurr when using an incorrect db backup.
```
jholdstock, line 58:
```suggestion
    Do **not** perform full-db backups unless you have very deep knowledge on how LN operates, how `dcrlnd` is implemented, and the _complete_ loss of funds that might occur when using an incorrect db backup.
```

docs/lightning-network/backups.md
```diff
@@ -0,0 +1,125 @@
...
+
+This transaction is updated and re-signed every time a new payment changes the relative balances of the channel. It is then stored on a local database, along with some other metadata such as the current state number, the specific keys used to sign this update and information about older, revoked states.
+
+A lightning node must **never** publish a state older than the most recent one agreed by the two parties of a channel. Due to how the LN protocol works, publishing the commitment transaction of an old, revoked state means that party forfeits their funds to the other node, who can then send the _entire_ channel amount to a regular key under their sole control.
+
+Notice that since the commitment data is never published to the blockchain during the regular operation of an LN node, any event that causes loss of data means the node cannot recover their funds if the counterparty disappears.
+
+It also means the node cannot even advance the state of the channel, since it doesn't know what the current state _is_ to begin with. 
+
+### Unsafety of copying the dcrlnd DB
```
jholdstock, line 52: This heading is not grammatically correct, but I'm not really sure of the best way to re-word it.

Perhaps "Danger of copying the dcrlnd database" ?

docs/lightning-network/backups.md
```diff
@@ -0,0 +1,125 @@
...
+
+In other words, using a backup with an old state causes _all funds to be sent to the remote party_. On the public blockchain, this looks as if the local node tried to steal funds (by broadcasting an older commitment tx) and the remote node then performed the "breach" protocol, exacting justice for the incorrect behavior of their counterparty.
+
+Therefore, trying to do a simple, continous backup of the `dcrlnd` DB is _unsafe_, in the sense that attempting to use such a backup risks all the funds in all of the node's channels.
+
+While it might be possible to offer some form of DB replication to increase reliability and safety for larger nodes and operators, this remains an open problem within the `dcrlnd` design and implementation.
+
+### The Data Loss Protect option
+
+The Lightning Network [BOLTs](https://github.com/lightningnetwork/lightning-rfc) define a special channel feature called `option_data_loss_protect`. Triggering this option during channel re-establishment (for example, after suffering data loss) causes the other node to publish their most recent (unrevoked) commitment transaction, thus in principle allowing a node to get their funds back.
```
jholdstock, line 75:
```suggestion
The Lightning Network [BOLTs](https://github.com/lightningnetwork/lightning-rfc) define a special channel feature called `option_data_loss_protect` (DLP). Triggering this option during channel re-establishment (for example, after suffering data loss) causes the other node to publish their most recent (unrevoked) commitment transaction, thus in principle allowing a node to get their funds back.
```

docs/lightning-network/backups.md
```diff
@@ -0,0 +1,125 @@
...
+
+However to retrieve those funds a node still needs some minimal metadata about a channel - mainly the specific keys and scripts used to redeem the local funds of the commitment transaction.
+
+In dcrlnd, this metadata and the mechanism to trigger restores is called the _Static Channel Backup_ (SCB) data.
+
+Note that since the DLP protocol involves asking the _remote_ node to close a channel due to some local data loss, the remote node could attempt to send an older state that is more to its advantage.
+
+Nodes are disicentivized to do this however, because the local node could be lying to force the triggering of the DLP (and thus the remote node publishing the old state allows the local node to send a justice transaction) and because the node could have previously set up a watchtower that could remedy any breaches attempted by their counterparty.
```
jholdstock, line 83: `disicentivized` missing `n`

This sentence is a bit too long, I think it should be broken down into two, something like:

> However, nodes are disincentivized to do this because the local node could be lying to force the triggering of the DLP, and if the remote node publishes an old state, the local node would be able to use its newer state to send a justice transaction which claims all of the channel funds.

And then a second sentence about watchtowers.

docs/lightning-network/backups.md
```diff
@@ -0,0 +1,125 @@
...
+
+In other words, an SCB file contains only a sort of _shell_ channel information: the minimum amount of info necessary to locate and redeem funds that belong to the local node on commitments created by and stored by the _remote_ party.
+
+Therefore, this file doesn't change during regular operations of a single, specific channel: the data in the SCB is valid for the entirety of the channel's lifetime.
+
+The SCB file is only updated once new channels are created (in which case the new channels are added to it) or closed (when a channel is finally removed from it).
+
+The contents of the file are also _encrypted_ using one of the node's private keys so they are only valid and decodable by the original node that created them.
+
+Note that SCB-based backups do not suffer from the unsafety limitations of the full DB copy procedure outlined previously. This is due to the fact that this method of backups does _not_ attempt to restore channel operations but only to trigger a remote channel closure via the DLP protocol so that a node may retrive its funds.
```
jholdstock, line 99:
```suggestion
Note that SCB-based backups do not suffer from the limitations of the full DB copy procedure outlined previously. This is due to the fact that this method of backups does _not_ attempt to restore channel operations but only to trigger a remote channel closure via the DLP protocol so that a node may retrive its funds.
```

jholdstock:
Thank you!

Issue #375, opened 2017-09-19
macOS appdata dirs should use uppercase app names


jrick:
They are all lower case at the moment (e.g. ~/Library/Application Support/dcrd).

While the HFS+ and APFS filesystems are case insensitive on macOS by default, they can be run in a case sensitive mode and the paths in the documentation will not work on these systems.

PR #1107, opened 2020-06-20
Wallets -> Decrediton -> Troubleshooting - fix macos config path

svitekpavel:
The path to Decrediton config is wrong, it is with lowercase `decrediton`.


jholdstock:
These were set to upper case for #375. @jrick could you comment on this one please?

jrick:
It's decrediton that's wrong (or rather, is different from all of the other decred tools):

```
% uname
Darwin
% ls ~/Library/Application\ Support/ | grep -iE '^(dcr|decred)'
Dcrctl
Dcrd
Dcrlnd
Dcrwallet
decrediton
```

But if decrediton isn't going to change this, then yes, the docs should be updated with the correct casing.

PR #1121, opened 2020-08-24
Update wallets and add mobile.

### LINKED ISSUE #1045: Mention mobile wallets
No mention of mobile wallets in the docs. 

A good place is here:
https://docs.decred.org/getting-started/beginner-guide/#wallets

A new section for mobile wallets and a tutorial similar to Decrediton may also be of value:
https://docs.decred.org/wallets/decrediton/decrediton-setup/

jholdstock:
Closes #1045

### DIFF
--- a/docs/getting-started/beginner-guide.md
+++ b/docs/getting-started/beginner-guide.md
@@ -2,17 +2,33 @@
 
 ---
 
-## Welcome To Decred 
+## Welcome To Decred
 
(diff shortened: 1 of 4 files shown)


fguisso:
According to https://decred.org/wallets/ there is more mobile wallets, I can't provide any reports, just for Exodus mobile that I use sometimes.

dajohi (approved):

PR #1127, opened 2020-09-22
Replace example dev proposal with a larger scale one

### LINKED ISSUE #905: Update Pi example proposal software feature
We recently made [updates to the proposal guidelines](https://github.com/decred/dcrdocs/issues/881). The marketing example still works, but the software feature example no longer reflects a typical Politeia proposal. It is smaller than any software proposals we've seen so far (it asks for $680 for ~2 weeks of work), and, as noted in this [issue](https://github.com/xaur/decred-issues/issues/33), Politeia is (currently) designed for larger proposals. Smaller proposals such as this would probably cost more to propose than implement. We don't want to say you can't submit proposals of this scale, but we should manage expectations and not promote small software proposals. Also, a small proposal doesn't provide the opportunity to illustrate a series of milestones (and drawdowns on those milestones).

jholdstock:
Addressing the issues raised in #905. Requesting review from @RichardRed0x and @xaur who have both previously demonstrated an interest in proposal guidelines and expectations.

As this is a example proposal it should really meet all of [the guidelines](https://docs.decred.org/governance/politeia/proposal-guidelines/) so I would appreciate some help/suggestions making sure this example is up to scratch.

If you hadn't guessed, I am using [this](https://proposals.decred.org/proposals/20e967dad9e7398901decf3cfe0acf4e0853f6558a62607265c63fe791b8b124) and [this](https://proposals.decred.org/proposals/ad0f968) as inspiration

Closes #905

### DIFF
--- a/docs/governance/politeia/example-proposals.md
+++ b/docs/governance/politeia/example-proposals.md
@@ -16,76 +16,101 @@ development of a new software feature. The second a marketing initiative.
 In the *What* section, provide a short description of the problem that will 
 be addressed.
 
-> Add monitoring RPC call that can be used for status reporting and as a heartbeat to ensure that politeiawww is functioning properly.
+> Develop a set of JavaScript libraries to enable JavaScript developers to
+> easily add support for Decred into their applications.
 
 ### Why
 
 In the *Why* section, answer why the proposal is needed and why it is
 a good idea.
 
-> Currently there is no prescribed way to remotely determine if politeiawww is functioning correctly. This is problematic because if there is a problem with politeiawww, the Politeia website could experience outages and Decred admins would not be aware of it until users were negatively impacted. We propose to add a single RPC that doubles as the heartbeat and a status monitor.
-> The idea is that monitoring software can periodically issue the RPC and determine if an alert needs to be sent out to the admins.
+> According to stats collected by GitHub.com, JavaScript has been the most
+> popular programming language for the last 6 years in a row.
+> The number of developers and companies building with JavaScript is enormous,
+> and providing a toolkit for them to easily integrate with the Decred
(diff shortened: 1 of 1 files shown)


xaur:
I like the idea to use an open demand (js lib) as an example. Might prompt someone to actually make such proposal, and then we can replace this example with the next relevant demand :)

The example reads well and it is a nice improvement over the existing one.

Not sure about using so much bold in When.

A thing to consider is the example rate of $30/h (down from $40/h in the current revision). Obviously it's just an example but it might be a starting point for dev proposals. Off top of my head $30 sounds quite low for the dev domain. I'm not familiar with real dev rates in Decred. What motivated this change?

Numeric bugs:

- the numbers in When add up to $14,820, not $14,520
- 40h for $1,500 is $37.5/h
- hours in 3 milestones add up to 444, not 484
- milestone 3 completes by week 14 while the total says 16

As an alternative to the bullet list you can try a table with right-aligned numbers (`--:` in the line between column header and first row).

Another thing is I'd like to demonstrate "continuity plan" in some example, as used in 3rd [bug bounty](https://proposals.decred.org/proposals/2170df6) proposal to avoid situations when proposal stalls when no fallback person is assigned. But it doesn't fit current and this new updated dev examples as they are single-person proposals. Perhaps we need another example that demonstrates a multi-person proposal.

jholdstock:
$30 was the result of me just picking a random number, I didn't really take notice of the numbers in the original example. I'll reset to $40 and fix the other numerical issues.

I like the idea to use a table, I will do that.

> hours in 3 milestones add up to 444, not 484

This is because 40 hours have already been completed.

xaur:
Looks good to me, thanks.

dajohi (approved):

Issue #1152, opened 2021-02-11
Use term StakeShuffle across the docs


xaur:
So it turns out that CoinShuffle++ is just one part of the equation. So far I've learned that

- what ends up on the blockchain is called **CoinJoin transactions**
- **CoinShuffle++** is a protocol and software to coordinate CoinJoins _in a way_ to make the server trustless. It could be some other way not CSPP at all. There is no such thing as CoinShuffle++ transactions and there is nothing CSPP on-chain.
- creation of CoinJoins is integrated with Decred's **Proof of Stake** system in a specific way

Recent marketing discussions suggest several people are happy with the StakeShuffle term to capture all of the above _and_ being less user-unfriendly at the same time. It was used in the [press release](https://decred.org/press/2021-01-25_three_new_features/)  and picked up by some news outlets (just google it). We'll be using it in Decred Journal too, unless a better term shows up and gains support.

Whatever the final term is, the incorrect usage of CoinShuffle++ is already out of the bottle. But at least we can fix some dcrdocs pages.

Places to update:

- [Overview](https://docs.decred.org/privacy/cspp/overview/) text
- [How to use CoinShuffle++](https://docs.decred.org/privacy/cspp/how-to-cspp/) text and title
- section name `Privacy / CoinShuffle++` (mkdocs.yml)
- when [Glossary](https://docs.decred.org/glossary/) is updated for privacy it should contain an entry for StakeShuffle (having an entry for CoinShuffle++ is ok as long as it is accurate on what it is)

dezryth:
I will work to address this in a PR for #1151 and #1152

xaur:
After recent [discussions](https://github.com/decred/dcrdocs/pull/1153#issuecomment-790614686), it must be clarified that this issue is about using term "StakeShuffle" in more "user facing" and "simplified" places (introductions, entry points that we share to newcomers), and it should be used as a synonym to "Decred Privacy", i.e. all elements involved.

CoinShuffle++ must still be mentioned where appropriate to give it due credit, along with other involved tech.

I think the right places to mention CSPP is anywhere where we need to break down what StakeShuffle is made of, some "advanced" pages, and places that talk specifically about configuring the CSPP server used.

Issue #1151, opened 2021-02-04
CoinShuffle++ Overview - Limitations: Section needs updating


dezryth:
I've identified an important area of the docs that need updating.

#### https://docs.decred.org/privacy/cspp/overview/#limitations
- Mixing is now available decrediton, but this section still suggests mixing is only available to command line dcrwallet and solo stakers. 
- There is a section of prerequisite bullet points that have now been met. This section should be reworded to describe the changes necessary and that they've been accomplished at this point, or be removed.

PR #1153, opened 2021-02-12
Update StakeShuffle/Privacy Documentation

dezryth:
This PR addresses #1151 and #1152. 

- The term StakeShuffle is used to describe Decred's implementation of the CoinShuffle++ protocol to create CoinJoin transactions. I attempted to replace the reference to CoinShuffle++ where StakeShuffle is more appropriate without making it too confusing as a lot of the URLs and documentation for command still references "cspp".
- I've updated verbiage in overview to reflect recent developments that have addressed limitations of the code since the docs were last updated. (VSPD allows mixing and staking/delegated voting rights, GUI is now available, regular non-staking transactions can be mixed)
- I've added a link to Exitus's privacy tutorial for participating in mixing via Decrediton on the "How to use StakeShuffle" page.

Please review and let me know of any changes anyone would like to see before merging. Given this is a very important area to Decred stakeholders, I want to ensure the verbiage changes are satisfactory.


raedah:
Can you update the urls and provide redirects for the old urls?

jrick:
Rebranding this feature is silly, and afaict, was not discussed or approved anywhere else.

The protocol is still CoinShuffle++ no matter how it is used and what amounts are mixed.  In fact, the server does not even know the current ticket price.  It only groups peers which are mixing the same amounts.

raedah:
The function of CSPP is distinct from coinjoins on chain. Due to the original branding of CSPP, people are frequently using the term incorrectly to describe the properties of the coinjoins on chain. CSPP is the client-server protocol for creating the coinjoin transactions. StakeShuffle was proposed as the branding for what people have been calling Decred Privacy. It makes a lot of sense since the primary usage is with the ticket system. The goal is to make this easier to understand for users and for terms to be used correctly.

dezryth:
I had some reservations about these changes myself that I'm more concerned with after jrick's response. CSPP is referenced a lot in the underlying code and documentation for the applications in their current versions. A rebranding needs wider consensus between developers and marketing. I think it may be best to revert most of the rebranding changes and perhaps just include a note that CSPP CoinJoin transactions used in combination with Staking have been referred to as "StakeShuffle". Calling it all StakeShuffle when it's not only ticket purchasing transactions that are mixed definitely seems off. This change introduces more confusion in it's current state. I will clean up and separate my changes into separate commits.

xaur:
Agreed @dezryth, if there is not enough consensus we need to discuss this further and undo the relevant changes here to not block this PR.

dezryth:
There has been more discussion on this topic and I feel more comfortable that this PR addresses the current takeaways I saw in [this discussion](https://matrix.to/#/!iCObfNbUZqNPUiRXxc:decred.org/$RxbXbR9dar6Jw9dWsIfkOHSFlq7g5BszT1tDlh2W890).

- Component CSPP++ needs to be mentioned/given credit in docs
- Most seem amenable to the overall process referred to as StakeShuffle 
- DCRWallet CLI commands etc will remain as is regarding cspp until a deprecation cycle 

I think my PR here accomplishes this actually, and I’m seeing more support than I expected for the branded name for the mixing process than I initially thought existed between marketing/dev. Can someone take a look at the changes here and let me know what parts in particular you take issue with in the review? I want to move it forward and this most recent conversation has been really helpful.

matthawkins90:
Great work! Looks good to me. I thought it might be valuable to link to the new glossary terms in the "Overview" section, but then I realized you had pretty much the exact same content covered there, so it would be redundant.

xaur:
I'd also replace the `cspp` in the URLs, but in a way that `cspp` links still work and redirects to the new ones.

PR #1160, opened 2021-03-16
mix a special unspent transaction output

zhizhongzhiwai:
dcrctl command used to mix UTXO


jholdstock:
I think this needs a little more explanation please, for example:

- What is meant by "special"?
- Why would this be necessary, what are the benefits?

PR #1164, opened 2021-04-25
add secure setup

degeri:


docs/advanced/secure-setup.md
```diff
@@ -0,0 +1,315 @@
...
+
+Now you can tar the  `copytob` folder for transport.
+
+`tar -zcvf ~/copytob.tar.gz -C ~/copytob decred-linux-{arm/amd}64-v{{ cliversion }}.tar.gz dcrd.conf rpc.cert`
+
+You can also calculate the hash to ensure that the file is not modified while moving.
+
+`sha256sum ~/copytob.tar.gz`
+
+Store this output on an uneditable medium. ie: paper or a photograph.
```
jholdstock, line 113: Why store this hash on paper/photo? Seems like a strange suggestion to me, I think we should justify it.

degeri: If we ask people to just store it people might save the hash into the computer itself. Which ofc can be manipulated by any malicious program.

docs/advanced/secure-setup.md
```diff
@@ -0,0 +1,315 @@
...
+## The Setup
+
+This will be a multi-system setup.
+
+![Network diagram](/img/secure-setup.png)
+
+* A general laptop or desktop that will be running the node. Referred as (`Computer A`)
+* A secure system that will be running our wallet for which we will be using a Raspberry Pi 4 device (It can be another laptop too).
+  Referred as (`Wallet-B`). This system should also be connected to an external monitor and keyboard.
+  It should also connect to the local network via ethernet.
```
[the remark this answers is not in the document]
degeri: Yes  .``Connected to Computer A via a crossover cable or a dedicated router`` sound better, lets go for that.

docs/advanced/secure-setup.md
```diff
@@ -0,0 +1,305 @@
...
+    [Download](https://github.com/decred/decred-binaries/releases/download/v{{ cliversion }}/decred-linux-amd64-v{{ cliversion }}.tar.gz) the Decred binaries for linux and place them in the same folder:
+
+    `wget -P ~/copytob https://github.com/decred/decred-binaries/releases/download/v{{ cliversion }}/decred-linux-amd64-v{{ cliversion }}.tar.gz`
+
+Make sure you [verify](verifying-binaries.md) the tar file to ensure it has not
+been tampered with.
+
+Now you can tar the `copytob` folder for transport.
+
+`tar -zcvf ~/copytob.tar.gz -C ~/copytob decred-linux-{arm/amd}64-v{{ cliversion }}.tar.gz dcrd.conf rpc.cert`
```
jholdstock, line 97:
```suggestion
`tar -zcvf ~/copytob.tar.gz -C ~/copytob decred-linux-{arm/amd}64-v{{ cliversion }}.tar.gz`
```

I believe this is wrong because `dcrd.{conf,cert}` are already in the copytob directory

degeri:
bump

docs/advanced/secure-setup.md
```diff
@@ -0,0 +1,314 @@
...
+
+---
+
+## The Setup
+
+This will be a multi-system setup.
+
+![Network diagram](/img/secure-setup.png)
+
+* A laptop or desktop that will be running the dcrd node, referred to as `Computer-A`.
```
matthawkins90, line 36: I think it would be better to immediately list the `Benefits of this Setup` section right after the picture. Readers who are "just checking it out" will probably be wondering "Why would I want to do this?" Also, the bullet points here don't necessarily explain the image above, so they feel disjointed.

My recommendation would be to delete these three bullet points, move directly into the `Benefits of this Setup` section below, and then go into further details about the hardware next.

docs/advanced/secure-setup.md
```diff
@@ -0,0 +1,314 @@
...
+  should also connect to the local network via ethernet.
+* All details/commands that come within curly braces `{}` cannot be copy-pasted,
+  you will need to remove the curly braces and edit them to suit your setup.
+
+### Benefits of this Setup
+
+* `Wallet-B` is kept offline and turned off most of the time.
+  This greatly reduces attack surface/opportunity.
+* An attacker would have to compromise dcrd on `Computer-A` first, and then find a
+  way to jump into `Wallet-B` using the RPC to compromise and exfiltrate data.
```
matthawkins90, line 49: Suggest changing this to:
"...and then find a way to break into Wallet B using the RPC to compromise and steal your wallet data. This is highly unlikely and difficult to pull off."

docs/advanced/secure-setup.md
```diff
@@ -0,0 +1,314 @@
...
+  This is highly unlikely and difficult to pull off.
+* dcrd on `Computer-A` can be used for other purposes in your local network (eg. other wallets, DEX, etc),
+  so you don't need to maintain multiple dcrd instances within your local network.
```
matthawkins90, line 52: Suggest adding two more bullet points as additional, benefits / explanation to this setup:

* Even though `Wallet B` is kept offline, it’s still possible for it to participate in staking by using a VSP.
* Advanced readers can extrapolate from this guide and make `Wallet B` a “hot” wallet so they can Solo Stake, but that’s significantly more vulnerable to attacks and outside the scope of this guide.

docs/advanced/secure-setup.md
```diff
@@ -0,0 +1,314 @@
...
+
```
matthawkins90, line 53: Before jumping straight into the instructions, I think it would be beneficial to add two more subsections here with a bit more preliminary explanation:

## Hardware

### Computer A - The Node

In general, you can use almost any computer to run the node. See the [Minimum Recommended Specifications](https://github.com/decred/dcrd#minimum-recommended-specifications-dcrd-only) for dcrd. `Computer A` is expected to remain online most of the time, but it's not required. For a budget device with a low energy requirement, it’s fine to use a Raspberry Pi with at least 2 GB memory and 16 GB disk space. It would also be fine to use a laptop or desktop. This guide assumes users are aware that, since Computer A is connecting to the public internet, then it is necessarily vulnerable to attacks.

### Wallet B

Since the wallet needs to have the utmost security in this setup, it’s important to have a minimal attack surface. This guide suggests that this wallet be a cold wallet: powered off most of the time, because it would be nearly impossible to attack a system which is powered off. This guide suggests using a Raspberry Pi for the wallet, although other hardware can be used.

### Finally, this guide suggests using a mouse, keyboard, and monitor to access 

## Operating Systems

Since `Computer A` is going to be the more vulnerable device, it doesn’t really matter which operating system you choose. `dcrd` can be installed on anything from Windows to Mac to Linux to OpenBSD. As mentioned before, you can just just use a laptop or desktop. However, if you plan to use a budget device like a Raspberry Pi that you can leave online, consider using Linux or OpenBSD, which will reduce your attack surface.

This guide assumes you already have an operating system on `Computer A`, and won’t describe how to install a new one.

`Wallet B` will be the more secure device. Use an operating system you feel comfortable with. Consider using Linux or OpenBSD to reduce your attack surface. Raspbian works fine if you want a GUI, and Ubuntu Desktop is another good choice for beginners who don’t need a GUI.

Again, this guide will give commands specific to Debian/Linux based systems. All details/commands that come within curly braces `{}` cannot be copy-pasted. You will need to remove the curly braces and edit them to suit your setup.

matthawkins90 (commented):
This guide is great, and a much-needed part of the core docs. I would love to see this added ASAP.

I haven't finished reviewing the actual instructions yet. However, I did some reorganizing and rewriting of the introductory segments.

I'm not a huge fan of the naming convention `Computer A` and `Wallet B`. It was confusing to me at first, because my mind immediately wondered if there was such a thing as a `Wallet A` that I needed to worry about. I would probably prefer something like `Computer A` and `Computer B`, or `Computer A - The Node` and `Computer B - The Wallet`, or simply just `The Node` and `The Wallet`. It's not a huge deal, though, so I continued using the same nomenclature you started with.

docs/advanced/secure-setup.md
```diff
@@ -0,0 +1,314 @@
+# <img class="dcr-icon" src="/img/dcr-icons/LockEye.svg" /> Secure Setup
+
+Last updated for CLI release v{{ cliversion }}.
```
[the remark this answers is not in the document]
jholdstock: Generally speaking, updating the whole site upon releases has been the approach thus far. A list of pages which should be reviewed upon release are included in README.md. I will add this one.

docs/advanced/secure-setup.md
jholdstock: > Since Computer A is going to be the more vulnerable device, it doesn’t really matter which operating system you choose. 

I don't follow the logic here. What do you think of something like

> Since Computer A is not storing any sensitive information, the choice of operating system is less important than the choice for Wallet B.

jholdstock:
Thoughts on renaming this guide to something more descriptive? eg `Cold Wallet staking setup`?

jholdstock:
Implemented most of your suggestions. Iagree with `Computer-A` and `Wallet-B` being kinda confusing. I will update to `The Node` and `The Wallet`.

docs/advanced/secure-setup.md
matthawkins90: Yep, good change.

docs/advanced/secure-setup.md
```diff
@@ -0,0 +1,372 @@
...
+```no-highlight
+mkdir ~/.dcrwallet/
+mkdir ~/.dcrctl/
+```
+
+Now, these files need to be edited.
+
+First copy the RPC username and password from ~/decredconfigs/dcrd.conf
+
+Let us name them {username} and {password}.
```
matthawkins90, line 298: This line should be changed.
I also think the username and password bit might deserve a tad bit more explanation, but I don't quite know the severity of messing up this step. Is the reader supposed to create their own username and password for these files, or modify them from the `~/decredconfigs/dcrd.conf`, or just copy them directly?

All I know is that I've set up several nodes and wallets on multiple isolated systems connected like this before, and this part is always confusing to me because I'm always concerned about which username and password to use, and if it's important that they be kept ultra-secret on the "vulnerable" node or not.

matthawkins90 (commented):
Awesome. It's coming together.

I haven't tried to reproduce the steps in the two walkthrough sections, so I haven't included that in my review. However, I've done all this process around three or four times, so I'm familiar with it all. It would probably be good for a final verification that all the exact steps make sense and are in a good sequential order.

The nice thing is, once we publish this and it goes live and people (especially novices) start testing it out, we'll start getting feedback about the process, and hopefully get good critiques to make it better.

My mindset is that, even though this guide kinda requires that people aren't afraid of terminals and installing Linux, it should still be aimed at being as accessible as possible to beginners who have no idea what they're doing. We *WANT* people to keep their Decred safe, and be happy with our software and community.

I think maybe one more review pass after these changes are made, and if @degeri has any feedback about the nomenclature of `Computer A` and `Wallet B` and is willing to update the .png, then we're almost there.

docs/advanced/secure-setup.md
```diff
@@ -0,0 +1,372 @@
...
+```no-highlight
+mkdir ~/decredconfigs
+mkdir ~/decred
+tar -xf /bconfig/copytob.tar.gz -C ~/decredconfigs
+tar -xf ~/decredconfigs/decred-linux-{amd/arm}64-v{{ cliversion }}.tar.gz -C ~/decred
+cp ~/decred/decred-linux-{amd/arm}64-v{{ cliversion }}/* ~/decred/
+rm -rf ~/decred/decred-linux-{amd/arm}64-v{{ cliversion }}/
+```
+
+Now we have the Decred binaries in the ~/decred/decred-linux-{amd/arm}64-v{{ cliversion }} folder.
```
matthawkins90, line 285: Possibly add a line here saying: "Don't run them yet. You need to make the config files first."
Also, do these binaries need to be added to a $PATH$ or something?

docs/advanced/secure-setup.md
```diff
@@ -0,0 +1,372 @@
...
+
+```no-highlight
+sudo ufw default deny outgoing
+sudo ufw default deny incoming
+sudo ufw allow out to {LOCAL_IP_OF_COMPUTER-A} port 9109
+sudo ufw enable
+```
+
+Replace {LOCAL_IP_OF_COMPUTER-A} with the IP address of `Computer-A`
+
```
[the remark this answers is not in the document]
matthawkins90: It would be especially good to have an example to compare against that even included a dumb, fake IP address like `192.168.1.50`, assuming that the IP of Computer A is that, that way we don't get people trying to enter commands like this:

`sudo ufw allow out to {192.168.1.50} port 9109`

and then asking why it didn't work

docs/advanced/secure-setup.md
```diff
@@ -0,0 +1,372 @@
...
+    Make a note of the generated hash.
+    For extra security, store the hash on an uneditable medium. e.g. paper or a
+    photograph.
+    A digital copy of the hash could be manipulated by any malicious program
+    running on the system.
+
+### Setting up Wallet-B
+
+Install an operating system you feel comfortable with.
+Ubuntu desktop is a good choice for beginners (there is no need for a GUI).
```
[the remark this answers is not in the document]
jholdstock: Ubuntu Desktop does have a GUI btw (hence the "desktop" name). Ubuntu server is the CLI only version.

docs/advanced/secure-setup.md
jholdstock: I agree that this can be a bit confusing but I think its a bit out of scope for this document. Would rather link to another page/external resource for more info.

jholdstock:
I've applied your latest suggestions @matthawkins90. I think this guide is looking great now, thanks a lot for pushing it over the line. Are you finished with your review?

matthawkins90 (commented):
Alright, sorry to do this to you, but I just had one more critique.
I went through all the instructions meticulously and tested a lot of it out looking for struggle points. I think after this we should be pretty great.

Otherwise, I'm satisfied with this guide and don't have any more feedback on it specifically. Looks great!

Next steps:
* Update the `{{ cliversion }}` var to make sure the links are all up to date.
* I feel as though some organization / internal-referencing of the docs would be beneficial. For instance, I realized we actually already have a "cool wallet" and a "hot wallet" guide under `/advanced/Solo-proof-of-stake-voting`, and we also have a `/advanced/running-a-full-node` guide as well, both of which seem pretty similar to this. Honestly, the main difference this guide offers is using two computers. I'm gonna try to see if I can some up with some good organizational changes.

README.md
```diff
@@ -97,6 +97,7 @@ $ docker run -d --rm -p <local port>:80 decred/dcrdocs:latest
 - `wallets/cli/cli-installation.md`
 - `advanced/solo-proof-of-stake-voting.md`
 - `advanced/verifying-binaries.md`
+- `advanced/secure-setup.md`
```
matthawkins90, line 100: Should be 'secure-cold-wallet-setup.md`

docs/advanced/secure-cold-wallet-setup.md
```diff
@@ -0,0 +1,390 @@
...
+
+1. Download Decred release binaries for `Wallet-B` and place them in the same folder.
+
+    === "Raspberry Pi"
+
+        ```no-highlight
+        wget -P ~/copytob https://github.com/decred/decred-binaries/releases/download/v{{ cliversion }}/decred-linux-arm64-v{{ cliversion }}.tar.gz
+        ```
+
+    === "Other Linux Systems"
```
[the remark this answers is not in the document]
jholdstock: Agreed

README.md
jholdstock: Good catch

jholdstock:
Pushed a commit to address the latest comments from @matthawkins90, and one more which re-formats the IP changes section so it is more consistent with the rest of the doc.

There is still space to improve this doc, but I think for now it is good enough to merge/deploy. We can polish it further in future PRs.

matthawkins90 (approved):

jholdstock (approved):

PR #1200, opened 2022-06-15
Use SVGs in issuance docs.

davecgh:
This updates the advanced issuance page to make use of SVGs for the images instead of PNGs since they scale better and are easier to modify if needed in the future.


jholdstock:
Increases the file size quite dramatically but not to the point of being an issue, worth it IMO. 

Also makes the image header brighter in dark mode, which is a nice improvement. It was definitely too dark before.

jholdstock:
Increases the file size quite dramatically but not to the point of being an issue, worth it IMO. 

Also makes the image header brighter in dark mode, which is a nice improvement. It was definitely too dark before.

davecgh:
> Also makes the image header brighter in dark mode, which is a nice improvement. It was definitely too dark before.

Yeah, I made the light mode image transparent and made the image header brighter in the dark mode image as you noted.

jholdstock (approved):

PR #1227, opened 2023-08-30
Upgrade build tools (mkdocs 9.2.6)

jholdstock:


davecgh (approved):

jholdstock:
Opened an issue on the mkdocs-material repo to hopefully understand why the build has started failing and if there is anything we can do about it: squidfunk/mkdocs-material#5924

PR #1235, opened 2023-11-10
Remove operating-a-vsp.md

jholdstock:
This will be moved into the vspd repo with other vspd specific doco.


jholdstock:
#decred/vspd/458

davecgh (commented):
What do you think about keeping a link somewhere to the documentation in the vsp repo?  While I agree it's not really reading for a general user, it is generally nice to leave breadcrumbs to help find content imo.

jholdstock:
How about this?

Screenshot from 2023-11-29 07-09-45

davecgh:
That looks good to me.

Issue #1237, opened 2024-02-07
FAQ section Decred: Where did it all begin? [link is expired]


exitus1:
No idea if temporary, but link/website is dead

https://docs.decred.org/faq/general/

https://thedecreddigest.org/2017/06/10/decred-where-did-it-all-begin/

PR #1236, opened 2023-12-14
Update mkdocs-material to 9.5.21.

jholdstock:
Includes lots of minor bug fixes, a smaller docker image, faster build times, and using a new emoji extension.

PR #1238, opened 2024-04-08
Fix thedecreddigest links (.org > .com)

pLabarta:
Solves #1237 and updates other broken links


jholdstock (approved):
Looks good, thanks.

Build should pass after #1236 is merged.

